PR #4, opened 2020-04-08
Refactoring for networkInstanceTest.

sadov:
Test that all 3 types (switch/L2, local, cloud) of the networks can be successfully deployed to an EVE instance.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -259,14 +259,42 @@ help:
 	@echo "This Makefile automates commons tasks of building and running"
 	@echo "  * EVE"
 	@echo "  * ADAM"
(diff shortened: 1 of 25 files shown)


rvs:
Hey @sadov -- in general we prefer smaller, self-contained PRs. For example, this one has Makefile logic changes mingled together with changes to the tests, it also seems to contain various stylistic changes (like changing the case of a variable name from upper to lower).

In the future, please make sure to split these types of commits into individual PRs, ok?

This time -- I'll merge it as-is.

sadov:
OK -- we will try to reduce the size of our PR.

PR #5, opened 2020-04-09
Tweaking Makefile to be more user-friendly

rvs:
Hey @sadov @giggsoff 

This PR makes Makefile more user-friendly, but more importantly it switches us to test with EVE KVM mode which is much, much, much faster than the default Xen one and allows default use of the ACCEL=y (did I mention it is much faster? 😃).

This, in turn, requires switch to the latest EVE release 5.1.10 -- but I hope this is OK, since you guys need to test with the latest anyway.

You can, of course, go back to the previous way of running things by specifying HV=xen and ACCEL=""

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,7 @@ $(BINDIR):
 
 EVE_DIST=$(DIST)/eve
 
+HV ?= "kvm"
 BIOS_IMG ?= $(EVE_DIST)/dist/$(ZARCH)/OVMF.fd
 LIVE_IMG ?= $(EVE_DIST)/dist/$(ZARCH)/live.img
 EVE_URL ?= "https://github.com/lf-edge/eve.git"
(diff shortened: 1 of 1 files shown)


sadov:
It seems that some changes have occurred in the naming scheme for kvm images -- perhaps we need to change our test of the base OS upgrade.

deitch:
Separately, I am beginning to wonder if we need to clone Adam at all. Adam already is a docker container [see here](https://hub.docker.com/r/lfedge/adam). Can we just run it? It also is a single binary, so we also could cut releases so all you need to do is `curl` it down.

Thoughts?

rvs:
> Separately, I am beginning to wonder if we need to clone Adam at all. Adam already is a docker container see here. Can we just run it? It also is a single binary, so we also could cut releases so all you need to do is curl it down.

I'd say not by default -- having that option available is useful tho (but as an option)

PR #8, opened 2020-04-09
Tweaking Makefile to be more user-friendly fix

sadov:
Some fixes of base image test for adoption of [PR5](https://github.com/lf-edge/eden/pull/5).

### DIFF
--- a/tests/integration/baseImage_test.go
+++ b/tests/integration/baseImage_test.go
@@ -25,6 +25,10 @@ func TestBaseImage(t *testing.T) {
 	if len(eveBaseRef) == 0 {
 		zArch = "amd64"
 	}
+	HV := os.Getenv("HV")
+	if HV == "xen" {
+		HV = ""
+	}
 	var baseImageTests = []struct {
 		dataStoreID       string
 		imageID           string
@@ -33,6 +37,7 @@ func TestBaseImage(t *testing.T) {
 		imageFormat       config.Format
 		eveBaseRef        string
 		zArch             string
+		HV                string
 	}{
 		{eServerDataStoreID,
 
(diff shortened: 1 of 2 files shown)


tests/integration/baseImage_test.go
```diff
@@ -25,6 +25,10 @@ func TestBaseImage(t *testing.T) {
 	if len(eveBaseRef) == 0 {
```
deitch, line 25: This file is a good example. We can let this change go through, but these tests in general should be in the eve repo, not here. When we want to change code for eve and its attendant tests, they should be in a single PR in a single repo.

rvs: 100% agreed @deitch ! In fact as I mentioned to @sadov yesterday -- we should freeze all new test development for now and focus on polishing the usability of the eden framework and test themselves. Fo that it makes sense to keep this repo monolithic for now to help with the speed of refactoring. However, once we are ready to start creating new tests -- I absolutely agree it would make sense to actually migrate them into the proper repos.

P.S. when I say "usability" it is things like this I have in mind https://github.com/lf-edge/eden/issues/10

tests/integration/baseImage_test.go
```diff
@@ -25,6 +25,10 @@ func TestBaseImage(t *testing.T) {
...
 		zArch = "amd64"
 	}
+	HV := os.Getenv("HV")
```
rvs, line 28: I'm fine with merging it for now, but we have to very quickly decide how to handle these types of config generically @sadov 

I filed https://github.com/lf-edge/eden/issues/10 to figure that one out -- please take a look

Ultimately, whether baseOS version contains a hypervisor label or not (and what kind) is completely orthogonal to what the test does. We have to have a way of making sure we can test any combination and I think that that can only be driven by external configuration.

PR #14, opened 2020-04-10
qemuconf added

sadov:
Yet another EVE QEMU running machinery with using of config file for -readconfig option.


Makefile
```diff
@@ -121,15 +119,40 @@ save: $(DIST)
...
+QEMU_ACCEL_N_Linux=-cpu SandyBridge,vmx=on
+QEMU_ACCEL:=$(QEMU_ACCEL_$(ACCEL:%=Y)_$(shell uname -s))
+QEMU_OPTS_amd64= $(QEMU_ACCEL)
+QEMU_OPTS_arm64= -machine virt,gic_version=3 -machine virtualization=true -cpu cortex-a57 -machine type=virt
+QEMU_OPTS_COMMON= -display none -serial mon:stdio -readconfig $(DIST)/qemu.conf -nodefaults -no-user-config
+QEMU_OPTS=-smbios type=1,serial=$(EVE_SERIAL) $(QEMU_OPTS_COMMON) $(QEMU_OPTS_$(ZARCH))
+
+QEMU_CONF_OPTS_amd64=
+QEMU_CONF_OPTS_arm64=--dtb-part=$(dir $(DEVICETREE_DTB))
+QEMU_CONF_OPTS=$(QEMU_CONF_OPTS_$(ZARCH))
```
rvs, line 150: I actually don't like this showing up directly in the Makefile. As I said -- the idea of hijacking EVE's makefile logic to run qemu is not the best one -- we should just own that part in Eden directly and we should do it through themplating.

Makefile
```diff
@@ -304,12 +333,11 @@ help:
 	@echo "   stop          stop ADAM and EVE"
 	@echo "   clean         full cleanup of test harness"
 	@echo "   eve-clean     cleanup of EVE instance related things"
-	@echo "   show-config   displays current configuration settings" 
+	@echo "   show-config   displays current configuration settings"
```
rvs, line 336: This seems like a spurious change (whitespace) -- please remove it

cmd/qemuconf.go
```diff
@@ -0,0 +1,332 @@
...
+
+type qemuDrive struct {
+	file      string
+	format    string
+	directory bool
+	label     string
+}
+
+func (drive *qemuDrive) toString() string {
+	output := fmt.Sprintln("[drive]")
```
rvs, line 39: Building qemu configuration like this bit by bit is doable, but brittle. What we should do instead is a templating approach of some kind. Here's how we do it in EVE itself: https://github.com/lf-edge/eve/blob/master/pkg/pillar/hypervisor/kvm.go#L40

And then https://github.com/lf-edge/eve/blob/master/pkg/pillar/hypervisor/kvm.go#L335

Please consider this approach instead.

Now, I'm actually not even saying it should be done based on Golang templates, given that it is highly likely that a user of Adam may want to tweak this config -- perhaps we can keep it externally as a template and then  decide whether we use golang templates or something like Mustache for actual templating.

@deitch may have other ideas here.

In fact, @deitch -- I am thinking that this could be useful in linuxkit itself -- basically providing an override qemu config for `linuxkit run qemu` use case -- what do you think?

deitch: For the first part, I get the appeal of doing it in go inside eden. Eden is a single command, and so being able to say, "give me qemu config" makes sense. But doing it bit by bit as rough, unless you are willing to write software that can manage the entire qemu config. However, to be fair, linuxkit does _precisely_ that, see [here](https://github.com/linuxkit/linuxkit/blob/master/src/cmd/linuxkit/run_qemu.go#L315-L337).

Is there any reason not to use all of those run options, or at least the struct? For that matter, we could externalize the private parts from linuxkit as a library so it can be leveraged elsewhere.

If that is what you want, raise an issue there.

rvs: I'd like to have the qemu configuration template that can be externalized just the same way we do it in EVE @deitch -- it is much easier to do small tweak that way to how qemu is run, as opposed to tweaking a ginourmous command line.

Also, I honestly really think that lk should have an option of consuming a fully formed qemu configuration file. May be ON TOP of what it does today -- but it would be very helpful.

PR #49, opened 2020-05-04
eve console

sadov:
Added `eden eve console` command.


sadov:
As I undersatnd it must be `eve-telnet-port`, but seems this paramenter is not added yet to pkg/utils/config.go for automatical config file generation. @giggsoff pls. look this point.

Issue #47, opened 2020-05-04
eden config generates incorrect eden.root


rvs:
My eden workspace is checked out at `/Users/roman/src/lfe/eden` however when I run `eden config` I see that the value of eden.root is pointing to `/Users/roman/eden`

sadov:
Fixed at [PR50](https://github.com/lf-edge/eden/pull/50)

mydatascience:
Likely not fixed. Eden config still tries to put config to /home/fleandr/.eden/ if I run from ~/eden2/eden

rvs:
Can we please get a resolution here @sadov @giggsoff ?

mydatascience:
I suggest to put /.eden/ in the current eden directory instead of home

mydatascience:
close? @rvs

Issue #42, opened 2020-04-29
We need to migrate away from file-based log and info gathering


rvs:
Currently Eden tries to follow logs and info messages that Adam collects through the filesystem. We need to migrate to something other than that and make sure that we can query Adam itself for that information. I'm sure @deitch will have an opinion here, but I'm thinking something like an embedded https://github.com/unit-io/tracedb into Adam.

What do you think @giggsoff @sadov -- can you guys articulate your requirements on this issue?

deitch:
I am going to draw out the picture as I understand it, both current state and desired. Let's ensure that it is correct, and then we can propose solutions.

### Current

Adam has a "driver" - currently supports memory and filesystem - where it stores information about devices, including (relevant to this discussion): info and logs.

eden's `log` and `info` commands (and their `*watch` variants) take an argument of a log file to parse. Essentially, `eden log <file>` is a smart log file parser; same for info.

To make this work you need to:

1. start the controller and device (e.g. via eden)
1. find the correct output file
1. pass the output file to eden via `eden log <filename>`

### Desired State

Independently of what we do here, the above functionally actually is quite useful, e.g. if we have a log file from elsewhere that we want to parse.

For normal eden functionality, though, we want this built into the inherent knowledge that eden has about running devices and controllers.

Our experience would look something like:

```console
$ eden log <device> [<args>]
```

Rather than needing to feed it the _file_, which we would have to find, we would just feed it the device ID, of which eden already is fully aware.

### Proposal

If the above is correct, I would propose the following:

1. Expose adam endpoints (when running as a server, of course) to get and/or stream logs and info for a device. This will abstract away the "file vs memory vs any other future driver" question, simplifying eden. It also will remove from the eden user the burden of finding the particular log file.
1. Construct eden to consume those adam endpoints. I would keep the file-based functionality we have now as an option; I really like the fact that I can sanely parse a log file using just eden, even without any running controllers or devices.

Once the above is done, I would look at embedding tracedb into adam. It wouldn't change anything for eden (at least initially), as it would be under-the-covers of adam. Eventually, we might improve the endpoints to make its usage more powerful. But one step at a time, methinks.

Does that sound about right?

rvs:
This is exactly right @deitch -- 100% agreed! Two additional points tho:
   * can you do a quick hack to at least expose raw files via http endpoint (e.g. just adding `http.FileServer(http.Dir(directory)` and hanging it off of an endpoint of some sorts?
[...]

Issue #22, opened 2020-04-17
Thoughts on EVE developer's workflow and how eden fits into that


rvs:
Hey @sadov @giggsoff @deitch @kalyan-nidumolu @eriknordmark I've started playing with Eden quite a bit this week and here's what emerged as my ideal developer workflow. Note that portions of it don't quite work with Eden yet, but if we all agree this is where we need to go -- we can then file individual issues to make it so.

For now, lets just use this issue to make sure we're all on the same page as to what that workflow should be.

The rest here is predicated on the following observations:
  1. EVE developers are focusing on hacking functionality that gets packaged via pkg/XXX
  2. ideally in my edit/compile/debug cycle I'd be able to simply update that one package in a running image of EVE someplace and continue with my testing -- I'm slowly working on making it a reality, but for now lets assume that EVE developers will always have to update a full image if they want to test their latest code in one of the packages
   3. regardless of how quickly we solve #2 in EVE, an average EVE developer doesn't really want to think about anything but rootfs they've just built -- all that business about installers, live images, etc -- that's deployment time. Most of my workflow is:
      3.1. edit packages
      3.2. compile rootfs
      3.3. debug it in a live version of EVE
 
If we all agree with the above, then it follows that the ideal workflow is having a semi-permanent running instance of EVE that we will keep "upgrading" (until it breaks or gets wedged -- and then we restart it to state 0 and keep "upgrading" again). 

Interestingly enough, this is exactly the kind of a workflow most of us who have access to commercial ZEDEDA's controller have been using for quite some time now. What is different now is that:
     1. we are trying to make the same workflow available to all the EVE developers
     2. we want to have a much, much finer grained control over the structure and the content of the config that EVE receives (speaking of which -- I would really love to see something like http://jsoneditoronline.org/ integrated directly into Adam, but it is fine for now)

So putting it all together, here's what my favorite workflow.

Before I begin my edit/compile debug cycle:
   * have a default `.eden.yml` checked into the EVE workspace - that would be a configuration file describing default way of deploying EVE and adam. You can either edit it or may be we can have an option of having `~/.eden/config.yml` that would always serve as an override. This file, but will allow me to tweak all the same value that today are in that make's Config.mk 
   * have a `~/.eden/state.yml` that keeps track of pairs of EVE and Adam running on this machine (I think it will be fair to assume for now that most of the time it'll be a single pair)
[...]

PR #50, opened 2020-05-05
New API and etc

sadov:
New EVE API, log/info from ADAM, rootfs download, get/set EVE config, binary tests with possibility of configuring by test script. Fix issues #[22](https://github.com/lf-edge/eden/issues/22), #[42](https://github.com/lf-edge/eden/issues/42), #[46](https://github.com/lf-edge/eden/issues/46) and #[47](https://github.com/lf-edge/eden/issues/47).


deitch:
Folks, I would love to review this, but it is too complex and big. There are 41 changed files, 21 commits, most of which are unrelated to one another. "use force-v1-api" is not related to "bin-test configuration" is not related to "added hooks support for tests".

Is there any reason this isn't 20 or so distinct PRs? Make them small and easy to review, we can get them merged in, and we run much less risk of breaking something. Granted, this is not in production use yet, but big PRs like this make it harder to get there, rather than easier.

Issue #45, opened 2020-05-03
We need to provide asynchronous asserts in Eden framework


rvs:
While looking at a few Eden tests it became clear to me that a we need to implement asynchronous asserts and in general encourage the test writers to write their tests with EVE's eventual consistency in mind.

Let me explain. Fundamentally, testing EVE is about the following 3 things:
   1. crafting a certain config based on the desired state that the test wants to validate
   2. monitoring EVE's state transition for valid (and invalid transitions)
   3. monitoring other state (outside of EVE) for desired properties

We are now starting to look good when it comes to automation of №1, and with log and info messages from Adam now moving to a continuous streaming architecture it is time we tackle №2.

The way I suggest we do it is -- we rely on two worker threads that always start when we initialize any controller/cloud object. Those worker threads are in business of continuously receiving log and info messages from the controller. What I'm suggesting is that we should make it possible to register listeners on these worker threads and make it convenient for test writers to create asserts that are attached to those listeners. Think of the asserts as validating (or invalidating!) the stream of info messages (and later logs).

E.g. instead of blocking like we currently do https://github.com/lf-edge/eden/blob/master/tests/integration/baseImage_test.go#L63 we will write a code similar to:

```
    deviceCtx.InfoAssertOnce("wait for version", filter.ZInfoDevSW, func(i InfoType) bool {i.version == desiredVersion})
```

This type of assert InfoAssertOnce is expected to fire once and when it does it is expected to go way (obviously it does have a chance to either pass or fail). At the end of the test run all these kinds of asserts that haven't passed are considered failed.

In addition we can also have a continuous type of assert -- the kind that simply listens and validates the events as they flow -- those types of asserts will persist until the end of the test run and then will have a chance to run one last time signaling the end of a testrun.

Obviously, we should also implement assert management that would allow us to block (if needed) on certain sets of asserts firing, etc.

I'd love to hear @deitch if he's seen any test frameworks like that -- if not -- it shouldn't be to difficult to implement from scratch.

Implementation-wise, I'm not sure whether all these asserts need their own goroutines and be fed events via channels or something else may be simpler/faster.

sadov:
Currently, we have some improvements (which we plan to place in the next PR) as an implementation of the functions described in the “dev. workflow” issue (https://github.com/lf-edge/eden/issues/22): to load the image of the base OS, send the JSON configuration to EVE and running the test binary with the` eden test` command.

And actually, we have a preliminary implementation of such asynchronous behavior:
[...]

PR #52, opened 2020-05-06
Tests sets support

sadov:
Some solution for issue #45
All tests in this tests set are executing by a separate goroutine, and all of them are completed after finishing of any test in this test set. The cumulative result of this test for the complete test set is just a result of this first finished test. Generally, you just need to create test functions and describe your test set (name, function, arguments and regular expression for the return value) in the ETests map. A simple example that you can see on `tests/integration/testset_test.go`

You can start them by eden from our test binary:
`./eden test -r TestSets/BoolTest -v debug`
or by traditional Go test procedure:
`go test tests/integration/testset_test.go tests/integration/tests_set.go tests/integration/common.go -run=TestSets/BoolTest -v`

The RebootTest test in this example will finish work after rebooting the EVE or timeout exceeding:
`./eden test -r TestSets/RebootTest -v debug`

### DIFF
--- a/tests/integration/tests_set.go
+++ b/tests/integration/tests_set.go
@@ -1,0 +1,65 @@
+package integration
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"strings"
+	"testing"
+)
+
+type EdenTestArgs map[string]string
+type EdenTestFunc func(t *testing.T, name string, args EdenTestArgs) string
+
+// EdenTest is a description of async test in tests set
+type EdenTest struct {
+	Name   string
+	Test   EdenTestFunc
+	Args   EdenTestArgs
+	Result string // Regexp pattern
+}
+
+type EdenTestSets map[string][]EdenTest
+
+// ETests is the main object for tests sets 
+var ETests EdenTestSets = EdenTestSets{}
+
+func runSubTest(t *testing.T, name string, test EdenTestFunc, args EdenTestArgs, result string, cntx context.Context, cancel context.CancelFunc) {
+	for {
+		select {
+		case <-cntx.Done():
+			return
+		default:
+			res := test(t, name, args)
+			matched, err := regexp.Match(result, []byte(res))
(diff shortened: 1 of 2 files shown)


rvs:
Hey @sadov -- after I merge your previous PR -- this one is now complaining "This branch cannot be rebased due to conflicts" -- can you please resubmit?

sadov:
Hey @rvs -- done.

tests/integration/tests_set.go
```diff
@@ -0,0 +1,65 @@
...
+        ctx, cancel := context.WithCancel(context.Background())
+        defer cancel() // cancel when we are finished tasks
+
+	name := strings.Split(t.Name(),"/")[1]
+	for _, tc := range ETests[name] {
+		fmt.Printf("Running %s.%s\n", t.Name(), tc.Name)
+		go runSubTest(t, tc.Name, tc.Test, tc.Args, tc.Result, ctx, cancel)
+	}
+	for {
+		select {
```
rvs, line 59: Why is this for here? @sadov

sadov:
Hey @rvs, I was thinking of a more abstract implementation of the arguments and results. But Implementing arguments as abstract interfaces may have some type checking problems, or we must implement test sets as methods for special classes. For return values, comparison with a regular expression for a string seems to be the most comprehensible comparison strategy, otherwise we need to define complex comparison functions (especially if we need to compare, for example, some fields in structures, slices or maps). And seems that representing of args/res as a string is closer to the possible configurations of text files. May be you have any ideas for better implementation? It would be great to discuss them.

tests/integration/tests_set.go
sadov: This is just a wait loop to avoid destroying the goroutines in a orphaned state after the parent function is completed. I have another implementation with a WaitGroup:
https://github.com/itmo-eve/eden/blob/tests_set_wg/tests/integration/tests_set.go

Issue #53, opened 2020-05-07
Add  timer.config.interval=5 to the initial config


rvs:
Hey @sadov @giggsoff to make sure tests run faster -- lets add the above to the initial config

sadov:
Fixed on [PR56](https://github.com/lf-edge/eden/pull/56)

Issue #51, opened 2020-05-06
Need to create an eden controller subcommand


rvs:
After some playing with Eden it has become clear to me that we need an `eden controller` subcommand that should be modeleted after zcli CLI API but capable of operating on the following controllers:
   * plain text files containing JSON EVE config
   * binary files containing protobuf serialization of the EVE config
   * Adam controller
   * Zedcloud

The way I envision this command is the following:

```
   eden controller -m [file|proto|adam|zedcloud]://<URL> [subset of zcli commands]
```

The semantics here is going to be that for plain text files, binary files and Adam controller the above command will simply take the source file from the -m option, parse it, apply the changes requested by zcli commands and store the result back (obviously in Adam's case that config needs to be retrieved via Adam's admin endpoint first and then submitted to that same endpoint -- but otherwise Adam use case is exactly like operating on a local JSON config).

To get us going, I suggest we only focus on plain text files with JSON config and Adam for now (but obviously the implementaion need to be designed in such a way that we can plug the other two easily later on).

And here's the first zcli command we need to implement `edge-node`. Hence the request here is for:

```
 edge-node reboot [-f] <name>
 edge-node update <name> [--title=<title>] [--lisp-mode=experimental|default] [--project=<project>] [--clear-onboarding-certs] [--config=<key:value>...] [--network=<network>...]
 edge-node eveimage-update <name> --image=<image> [--activate]
 edge-node eveimage-remove <name> --image=<image>
```

deitch:
I am not sure I understand. Is the goal here to be able to tell a controller to update the config for a device, etc.? i.e. tell it to take actions?

What is the `file` or `json` or whatever mode for? Or do these really mean, "I would want to change the config for a device to tell it to do X, but rather than doing it manually, I want a convenient CLI that updates a standard config"?

rvs:
Hey @deitch 

> I am not sure I understand. Is the goal here to be able to tell a controller to update the config for a device, etc.? i.e. tell it to take actions?

Correct. The way to think about it is this: at the end of the day the raison d'être of any controller is to build a config for EVE. That's where everything bottoms out. Controller builds that config based on the actions that are performed via API calls of controller itself. Those API calls are wrapped in zcli commands that we're modeling here.

[...]

PR #56, opened 2020-05-12
Controller command

sadov:
Implemented commands from [issue 51](https://github.com/lf-edge/eden/issues/51) for `file` and `adam` locations. The '&lt;name&gt;' is not implemented -- used single EVE instance from EDEN's config. For 'edge-node update' implemented only `--config=<key:value>` options.

Fixed [issue53](https://github.com/lf-edge/eden/issues/53).

### DIFF
--- a/cmd/changers.go
+++ b/cmd/changers.go
@@ -1,0 +1,105 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/lf-edge/eden/pkg/controller"
+	"github.com/lf-edge/eden/pkg/controller/adam"
+	"github.com/lf-edge/eden/pkg/device"
+	"github.com/lf-edge/eve/api/go/config"
+	"github.com/spf13/viper"
+	"io/ioutil"
+	"os"
+	"strings"
+)
+
+type configChanger interface {
+	getControllerAndDev() (controller.Cloud, *device.Ctx, error)
+	setControllerAndDev(controller.Cloud, *device.Ctx) error
+}
+
+type fileChanger struct {
+	fileConfig string
+}
+
+func (ctx *fileChanger) getControllerAndDev() (controller.Cloud, *device.Ctx, error) {
+	if _, err := os.Lstat(ctx.fileConfig); os.IsNotExist(err) {
+		return nil, nil, err
+	}
+	var ctrl controller.Cloud = &controller.CloudCtx{Controller: &adam.Ctx{}}
+	data, err := ioutil.ReadFile(ctx.fileConfig)
+	if err != nil {
+		return nil, nil, fmt.Errorf("file reading error: %s", err)
+	}
+	var deviceConfig config.EdgeDevConfig
+	err = json.Unmarshal(data, &deviceConfig)
+	if err != nil {
+		return nil, nil, fmt.Errorf("unmarshal error: %s", err)
+	}
+	dev, err := ctrl.ConfigParse(&deviceConfig)
+	if err != nil {
+		return nil, nil, fmt.Errorf("configParse error: %s", err)
+	}
+	return ctrl, dev, nil
+}
+
+func (ctx *fileChanger) setControllerAndDev(ctrl controller.Cloud, dev *device.Ctx) error {
+	res, err := ctrl.GetConfigBytes(dev)
+	if err != nil {
+		return fmt.Errorf("GetConfigBytes error: %s", err)
+	}
+	if err = ioutil.WriteFile(ctx.fileConfig, res, 0755); err != nil {
+		return fmt.Errorf("WriteFile error: %s", err)
+	}
+	return nil
+}
+
+type adamChanger struct {
+	adamUrl string
+}
+
+func (ctx *adamChanger) getControllerAndDev() (controller.Cloud, *device.Ctx, error) {
+	ipPort := strings.Split(ctx.adamUrl, ":")
+	ip := ipPort[0]
+	if ip == "" {
+		return nil, nil, fmt.Errorf("cannot get ip/hostname from %s", ctx.adamUrl)
+	}
+	port := "80"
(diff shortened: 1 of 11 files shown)


pkg/controller/device.go
```diff
@@ -45,6 +46,143 @@ func (cloud *CloudCtx) StateUpdate(dev *device.Ctx) (err error) {
...
+		return uuid.Nil, fmt.Errorf("nil UUIDandVersion")
+	}
+	if uuidAndVersion.GetUuidandversion().GetUuid() == "" {
+		return uuid.Nil, fmt.Errorf("nil UUIDandVersion")
+	}
+	return uuid.FromString(uuidAndVersion.GetUuidandversion().GetUuid())
+}
+
+//ConfigParse load config into cloud
+func (cloud *CloudCtx) ConfigParse(config *config.EdgeDevConfig) (device *device.Ctx, err error) {
```
rvs, line 83: small nit: would be useful to be super explicit that this function not so much parses config (there's nothing to parse -- it is all in go structures in config.EdgeDevConfig) but rather populates or builds your internal representation of config from protobuf defined go structure.

So at least lets reflect that in the function's name. On top of that, is there any reason not to use the actual config.EdgeDevConfig as the basis of device.Ctx? Along the lines of:
```
struct device.Ctx {
    config.EdgeDevConfig
}
```

?

cmd/changers.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	return nil
+}
+
+type adamChanger struct {
+	adamUrl string
+}
+
+func (ctx *adamChanger) getControllerAndDev() (controller.Cloud, *device.Ctx, error) {
+	ipPort := strings.Split(ctx.adamUrl, ":")
```
rvs, line 62: I don't think parsing of the CLI arguments belongs in the utility function -- it probably belongs to adamChanger context that neeeds to be populated from the actual cmd line function. Lets refactor it that way so that these functions simply operate on well-populated context objects.

Makes sense?

cmd/changers.go
```diff
@@ -0,0 +1,105 @@
...
+	if ip == "" {
+		return nil, nil, fmt.Errorf("cannot get ip/hostname from %s", ctx.adamUrl)
+	}
+	port := "80"
+	if len(ipPort) > 1 {
+		port = ipPort[1]
+	}
+	viper.Set("adam.ip", ip)
+	viper.Set("adam.port", port)
+	ctrl, err := controller.CloudPrepare()
```
rvs, line 73: Apropos previous point -- I really don't want to see viper.* calls in the utility functions.

cmd/edenController.go
```diff
@@ -0,0 +1,463 @@
...
+	"strings"
+)
+
+var controllerMode string
+var baseOSImage string
+var baseOSImageActivate bool
+var configItems map[string]string
+
+func getParams(line, regEx string) (paramsMap map[string]string) {
+
```
rvs, line 24: Why are we not simply using net/url.Parse?

cmd/edenController.go
```diff
@@ -0,0 +1,463 @@
...
+	if len(params) == 0 {
+		return "", "", fmt.Errorf("cannot parse mode (not [file|proto|adam|zedcloud]://<URL>): %s", controllerMode)
+	}
+	ok := false
+	if modeType, ok = params["Type"]; !ok {
+		return "", "", fmt.Errorf("cannot parse modeType (not [file|proto|adam|zedcloud]://<URL>): %s", controllerMode)
+	}
+	if modeURL, ok = params["URL"]; !ok {
+		return "", "", fmt.Errorf("cannot parse modeURL (not [file|proto|adam|zedcloud]://<URL>): %s", controllerMode)
+	}
```
rvs, line 48: Same question about using net/url.Parse

cmd/edenController.go
```diff
@@ -0,0 +1,463 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		modeType, modeURL, err := getControllerMode()
+		if err != nil {
+			log.Fatal(err)
+		}
+		log.Infof("Mode type: %s", modeType)
+		log.Infof("Mode url: %s", modeURL)
+		var changer configChanger
+		switch modeType {
+		case "file":
```
rvs, line 90: So as per my comment above the right place to hook up the parsing of viper options is here and then we can pre-populate the *Changer context objects with what they have to know.

cmd/edenController.go
```diff
@@ -0,0 +1,463 @@
...
+		if err != nil {
+			log.Fatalf("GetFileFollowLinks: %s", err)
+		}
+		modeType, modeURL, err := getControllerMode()
+		if err != nil {
+			log.Fatal(err)
+		}
+		log.Infof("Mode type: %s", modeType)
+		log.Infof("Mode url: %s", modeURL)
+		var changer configChanger
```
rvs, line 144: Feels like these repeated blocks need to be refactored down to a utility method so we don't have code duplication between all these different viper commands.

cmd/edenController.go
```diff
@@ -0,0 +1,463 @@
...
+		default:
+			log.Fatalf("Not implemented type: %s", modeType)
+		}
+
+		ctrl, dev, err := changer.getControllerAndDev()
+		if err != nil {
+			log.Fatalf("getControllerAndDev error: %s", err)
+		}
+
+		dataStore := &config.DatastoreConfig{
```
rvs, line 160: Maybe this is a bit premature, but it seems like we need to wrap all those config manipulations into SDK functions of some kind -- one option of doing that by using Go's WithOptions design pattern: https://www.sohamkamani.com/golang/options-pattern/

rvs:
Btw @sadov -- now that I'm merging it -- let me know how do you want to track refactoring that I suggested. Filing GH issues or something else?

Issue #41, opened 2020-04-28
eden status needs to provide much more detailed information on various artifacts


rvs:
Currently, running `eden status` prints a pretty brief summary covering only "running entities" (Adam, EServer and EVE itself):

```
Adam status: adam container not exists
EServer status: no pid file
EVE status: no pid file
```

We need to extend its functionality to cover all the artifacts that can be mentioned in the config and their state.

mydatascience:
Now eden status reflects only whether something is running or not.

Suggested:
If something is running display (except name & status):

IP/Port (how to connect) 2) log file location
If something is not running display (except name & status):
Likely cause (if possible) 2) log file location 3) last error from log (something like tail -f)

mydatascience:
Currently an EVE instance under Eden's management can be in one of the following states:

unknown
onboarding
managed
These would correspond to the status of the adam admin device/onboard commands and need to be displayed by eden status #37

mydatascience:
root@NPC:~/eden# ./eden status
EVE state: registered

Adam status: container with name eden_adam is running
	Adam is expected at https://192.168.1.241:3333
	For local Adam you can run 'docker logs eden_adam' to see logs
Redis status: container with name eden_redis is running
	Redis is expected at redis://192.168.1.241:6379
	For local Redis you can run 'docker logs eden_redis' to see logs
EServer process status: running with pid 26039
	EServer is expected at http://mydomain.adam:8888 from EVE
	Logs for local EServer at: /root/eden/dist/eserver.log
EVE process status: running with pid 26049
	Logs for local EVE at: /root/eden/dist/eve.log 

LGTM. Close?

PR #81, opened 2020-05-30
Improved 'eden status' command output

sadov:
'Status' related part of splited https://github.com/itmo-eve/eden/pull/11


mydatascience:
for #41

PR #116, opened 2020-06-24
Eden test fix

sadov:
Tests scenario args added.


rvs:
Why was this closed @sadov ?

Issue #104, opened 2020-06-18
Usage of asynchronous asserts is still overly complicated


rvs:
If for a simple test reboot counter we have to write this much extra code https://github.com/lf-edge/eden/blob/master/tests/reboot/check_reboot.go clearly things can be simplified.

I'd like to suggest we take a long step back and start from the very basics. Mainly, lets just start with implementing TestContext.AddProcInfo, TestContext.AddProcLog and TestContext.AddProcMetric -- all of these functions will take an argument of a Processor function that simply accepts individual entries of Info, Log and Metric type, does whatever it needs to do with them and if it returns `err != nil` gets deregistered (for as long as it returns nil it keeps getting called).

That way we can have, effectively, an open-ended stream processing of the 3 streams and we can figure out what else we can build out of that step-by-step.

mydatascience:
We have this https://wiki.lfedge.org/display/EVE/Objects+and+methods+for+SDK and now refactoring the code, hiding typical AssertFunctions in the SDK. 
Generally agree to make it much more simple

PR #118, opened 2020-06-25
Log/Info/Metric tests

sadov:
Related to https://github.com/lf-edge/eden/issues/104.

Added Logs/Metrics support for Process/State machinery. Improved 'eden test` functionality for passing parameters to scenario tests. Log/Info/Metric tests are implemented, which can be performed, for example, with the following command:
```
./eden test tests / lim / -v debug -a '-timewait 600 -number 3'
```
In this case, we will run TestLog, TestInfo and TestMetrics with the expectation of 3 events (by this parameter we check the deactivation of the Process function after a non-zero return, by default we expect 1 event) and 10 minutes. timeout (test for asynchronous processes, default 1 min.)

### DIFF
--- a/cmd/edenTest.go
+++ b/cmd/edenTest.go
@@ -1,12 +1,12 @@
 package cmd
 
 import (
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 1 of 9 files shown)


rvs:
LGTM. I'll be merging this, but we need to really cleanup Makefile infrastructure if we intend to keep using it. There's WAAAY too many repetitions there.

PR #131, opened 2020-06-30
fix EVE naming

sadov:
Fix. for https://github.com/lf-edge/eden/issues/123#issuecomment-651386687 2.


mydatascience:
This is to be able to refer to EVE name even if EVE was onboarded using the command `eve onboard`. The problem is that EDEN doesn't send some parts of the config and ssh is lost

PR #133, opened 2020-06-30
ssh key set on every eden eve ssh

sadov:
Fix for https://github.com/lf-edge/eden/issues/123


mydatascience:
We want to add the ssh key to ssh access. Before that - we may have assumed that ssh was set up by someone else and is unchanged. It's a wrong assumption

PR #137, opened 2020-07-03
Fixes in Makefiles

mydatascience:
This PR is about improving `make` command for tests. 
Before make contained building separate tests explicitly. Now we build everything we find in tests folder. 

Also we allow CONFIG to be specified  - this allows to have a one-liner for running tests on non-Qemu config. 
`make CONFIG='--devmodel RPi4' run`
We add TESTS param to specify which tests to use 
`make TESTS='lim units' test`
And a DEBUG param

Also we clean some obsolete messages about uuidgen


sadov:
@rvs Actually, these changes were inspired by your comment: https://github.com/lf-edge/eden/pull/118#issuecomment-649937446 and we now have a smaller Makefile without the need to change while adding a new test. Tnanks for your comment.

Parameters have been added because we must have possibility for choose specific tests from autogenerated list for `make test` and we will have a lot of problems compiling exec- and test-binaries by `make build`. Prior to this, the default configuration generated only for the default model - ZedVirtual-4G, and it was necessary to make several magic passes to build a complete set for Rpi, for ex. No we

Issue #135, opened 2020-07-02
make eden eve reset command


mydatascience:
Make a command that allows to restore eve state to the currrent config. 
This will allow to restore corrupted config (e.g. with ssh/wifi)  if eve still can connect to Adam

PR #138, opened 2020-07-03
Eden eve reset command and onboard fix 

mydatascience:
We introduce `eden eve reset` command that allows to reset the config to the initial one (global) that is set up by `eden config add` and modified manually. This is useful if something bad happens and config is broken, but connection between EVE and ADAM is still available e.g. Raspberry lost WiFi or ssh or just you want to restore the environment.

Also there is a fix for `eden eve onboard` when eve is onboarded after the timeout of` eden eve onboard` command
#135


rvs:
LGTM!

PR #161, opened 2020-07-16
CircleCI for building & publishing  EServer container

mydatascience:
This is a placeholder for the automation via CircleCI and publishing the image of eserver to Dockerhub
We need to deal with image tags though. So maybe this PR will be updated


mydatascience:
Moved to Github Actions. Bye CircleCI

PR #167, opened 2020-07-16
Update Readme to include a how-to get an EVE version and syslog from SD card

mydatascience:
Included updated Readme. 


rvs:
I'll be merging this now @mydatascience -- but see my comment above

PR #201, opened 2020-08-06
Tracking state of the application

mydatascience:
We add this application test state to support this functionality in further tests and test that we can correctly use messages to know the actual state. Alternative is use `eden pod ps `


mydatascience:
will not use for now

PR #207, opened 2020-08-11
add lots of docs

deitch:
As discussed on Slack with @giggsoff


deitch:
This need not be perfect and complete, just an improvement. If it is incomplete, let's get it in; if something is incorrect, let's fix it first.

PR #224, opened 2020-08-17
provide option to override uefi tag

### LINKED ISSUE #223: support separate tags for eve and eve-uefi
eve-uefi is rebuilt rarely, especially during an eve dev cycle, while eve is rebuilt regularly. To boot, eve-uefi is slow and expensive to build.

Currently, we use the same tag for eve and eve-uefi. We should have the ability to override it.

deitch:
Fixes #223 

When working with developing eve, you need to change and rebuild it frequently. You do _not_ want to have to rebuild UEFI all of the time. This lets you override just the eve tag, while keeping the uefi tag the same.

### DIFF
--- a/cmd/downloader.go
+++ b/cmd/downloader.go
@@ -2,19 +2,21 @@ package cmd
 
 import (
 	"fmt"
+	"path/filepath"
+	"runtime"
+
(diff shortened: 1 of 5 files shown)


rvs:
LGTM

PR #241, opened 2020-08-30
Nice status indicators

deitch:
Just for ease of instant viewing.


deitch:
Also, while I was at it, fixed a log message type that said "trying to SHH" to "trying to ssh"

PR #254, opened 2020-09-09
add local docker registry

deitch:
In addition to eserver, this starts up a local docker registry, which the eve can use as a datastore. It adds it to `eden status` and `start` and `stop`, and provides an `eden registry` command, which has:

* `start`
* `stop`
* `status`
* `load`

The latter is interesting. `eden registry load` will look for an image, first in your local docker cache (`docker image ls`) and then, if it cannot find it, in whatever registry is appropriate for the provided image name (exactly how `docker run` does it). It will take that image and load it up into the local registry.

Finally, you can tell eve to run a pod from the local registry by using:

```console
eden pod deploy --registry=local docker://nginx
```

README has been updated.

Almost all of the parts have been tested and function.

The functionality telling it to deploy from a local registry (`eden pod deploy --registry=local docker://nginx`) is not yet implemented. This was complex enough that it made sense to get this in and working, and add that as a separate PR.

cc @rvs


rvs:
I love this -- great work @deitch! One quick question though -- since we started moving into the direction of magic options, wouldn't a magic URL work as well? E.g. `docker://local/library/nginx` or something like that? For me it would be more regular usage.

deitch:
I thought about it but couldn't come up with a sane one. Anything that changes the url felt too weird.

rvs:
Fair enough @deitch -- @giggsoff @mydatascience any feedback before I merge this?

mydatascience:
looks good to me but I believe Roman's suggestion is a good option. docker://local/nginx
Also I suggest to have it as an option. I'd prefer not to have even redis locally, so eden should be a lightweight thing. 
Imagine you have adam in the cloud, docker registry and eserver in the cloud and eden on the local machine. 
Eden might be even on some edge device which can power off your robots )

deitch:
Two comments:

- The actual pull from local isn't properly implemented yet, so deciding on the format isn't so important for this PR and can be covered separately
- Agreed on possible having redis and registry as an option. Neither is now, so we should take this as is, and then add options to both

rvs:
Fair enough @deitch -- can you please resolve conflicts so we can pull this?

deitch:
> can you please resolve conflicts so we can pull this?

Done.

PR #321, opened 2020-10-15
Workflow test variants.

sadov:
The predefined workflow test script eden.workflow.tests.txt supports the EDEN_TEST environment variable for setting of test variants:
* The default variant without setting the 'EDEN_TEST' variable -- optimized for the usual test suite of CI / CD tasks. You need to reset/clean `eden` before running.
* Variant 'small' -- optimized for quick testing of coomon development workflow. No cleaning and reinitialization required between test cycles.
* Variant 'large' -- complete set of tests for full testing. You need to reset/clean `eden` before running.

The test variant can be selected at runtime by setting the environment variable:
```
EDEN_TEST=small eden test tests/workflow/
```
or by chosing of autogenerated by `eden` template machinery workflow scenario:
* `workflow.default.txt`
* `workflow.small.txt`
* `workflow.large.txt`

For example:
```
eden test tests/workflow/ -s workflow.large
```


mydatascience (changes requested):
I feel it's better to have several versions of workflow. Also I'd like to add the option to be able to start a workflow from some tagged line. It looks it will be too complicated to handle and modify if we accept this PR as is.
Let's have 3 eden.workflow.tests.txt and one command (might be as it is now)
Like  small.workflow.tests.txt medium.workflow.tests.txt large.workflow.tests.txt? 
I'd really want to keep some very very simple file with workflow config. Now eden high-level scenarios serve it well.

rvs:
LGTM

PR #325, opened 2020-10-20
updated workflows 

mydatascience:
Set several types of workflows:
large
small
default


rvs (commented):
LGTM

PR #337, opened 2020-10-26
New network tests

sadov:
Created universal `eclient` container with SSH access. Old network test split into two tests: simple `nc`-based client/server communication and `nginx` tests.


mydatascience (changes requested):
We do not need this file at all 
These ports should be set by test user

sadov:
Which file?

mydatascience:
tests/eclient/eden+eclient.sh  

This one. This one adds port and restarts eden 

We require a user to explicitly define ports upon start of the test. We don't want start/stop eden within the test.

sadov:
It will not stop/start `eden` if the ports are already set. We will need this script to run the test from scratch.

rvs (approved):
LGTM

PR #338, opened 2020-10-26
Mariadb test

sadov:
`Eclient` docker based test for standard `mariadb` container.


.yetus-excludes
```diff
@@ -0,0 +1 @@
+.*\.txt
```
mydatascience, line 1: We don't want that. 
Why?

tests/eclient/eden+eclient.sh
```diff
@@ -7,22 +7,11 @@ EDEN=./eden
 if $EDEN config get --key eve.hostfwd | grep -v 2223
```
mydatascience, line 7: This file should not be here at all as in any other test

.yetus-excludes
sadov: Because `yetus` make a lot of crazy checking about tabs in escripts. But we need tabs in this or may be another test, just because we use `mysql` output with tabs to check the test results.

tests/eclient/eden+eclient.sh
sadov: https://github.com/lf-edge/eden/pull/337#issuecomment-716670577

.yetus-excludes
rvs: the we should subset it to just escripts folders -- disabling it everywhere seems too agressive

.yetus/blanks-tabs.txt
```diff
@@ -1,3 +1,4 @@
 .*Makefile.*
 .*\.go
 .*\.md
+.*\.txt
```
rvs, line 4: ditto

.yetus/excludes.txt
```diff
@@ -1,2 +1,3 @@
 ^.+go.mod
 ^.+go.sum
+.*\.txt
```
rvs, line 3: ditto

rvs (approved):
A few yetus related nits -- LGTM otherwise

.yetus-excludes
sadov: OK -- `yetus` check is now disabled only for MariaDB escript.

mydatascience: Fine

PR #342, opened 2020-10-29
Host-only ACL test

sadov:
Just ping's from common and host-only application containers.


rvs (commented):
LGTM

PR #358, opened 2020-11-10
parallels configs

dautovri:
Critical configs to run Eden on Linux on parallels(MacOS)


README.md
```diff
@@ -498,3 +500,9 @@ If you want to clean artifacts of all contexts, you should to run `clean --curre
 * `eden utils gcp` - sub-commands to work with Google Cloud Platform
 * `eden utils export` - sub-command to save certs and configs into tar.gz
 * `eden utils import` - sub-commands to load certs and configs into tar.gz
+
+## Parallels
+
+0. Shut down the virtual machine before changing the hypervisor type.
+1. Open virtual machine configuration > **Hardware > CPU & Memory > Advanced Settings**, and set Hypervisor type to **Parallels**.
+2. Click the **Enable Nested Virtualization** checkbox.
```
mydatascience, line 508: The description is not enough. Please provide a detailed one. 
You need to consider somebody who wants to use eden +eve on Mac with ability to launch VM's in EVE. 
Your instruction is not clear.  Also you may want to add pictures

Should start like: 
If you want to be able to launch VMs or containers-in-VMs or accel=true on Mac 
You need to use Parallels. 
1) Set up parallels ... 
2) 
3)

mydatascience (changes requested):

README.md
```diff
@@ -498,3 +500,9 @@ If you want to clean artifacts of all contexts, you should to run `clean --curre
 * `eden utils gcp` - sub-commands to work with Google Cloud Platform
 * `eden utils export` - sub-command to save certs and configs into tar.gz
 * `eden utils import` - sub-commands to load certs and configs into tar.gz
+
+## Parallels
+
+0. Shut down the virtual machine before changing the hypervisor type.
```
rvs, line 506: Wait, wait -- where's the virtual machine coming from? Seems like some steps are missing right before this one.

rvs (changes requested):
Hey @dautovri -- this is a good start, but what we really need is a support for Parallels CLI inside of EVE's own Makefile infrastructure.

Parallels comes with cli utilities `prl_convert    prl_disk_tool  prl_perf_ctl   prlcore2dmp    prlctl         prlexec        prlsrvctl` which can be used in place of qemu (with nested virtualization enabled). It would be REALLY cool to first implement support for those in EVE's own Makefile and then also add support to Eden.

E.g. my ideal workflow is to be able to either type `make ACCEL=parallels run` in EVE's workspace or `eden config add default --model parallels` in the Eden workflow.

Do you think you can do it as the follow up to this README update?

dautovri:
> Parallels comes with cli utilities `prl_convert prl_disk_tool prl_perf_ctl prlcore2dmp prlctl prlexec prlsrvctl` which can be used in place of qemu (with nested virtualization enabled). It would be REALLY cool to first implement support for those in EVE's own Makefile and then also add support to Eden.
> 
> E.g. my ideal workflow is to be able to either type `make ACCEL=parallels run` in EVE's workspace or `eden config add default --model parallels` in the Eden workflow.

@rvs @mydatascience Thank you for your suggestions. I will figure out how to make it and will integrate parallels in Makefile.

rvs (commented):
OK, I'm confused now -- the last picture here shows that we're running Ubuntu -- I thought the whole point of the exercise was to run EVE under parallels. Am I missing something @dautovri @mydatascience  ?

PR #367, opened 2020-11-16
EDEN's port manipulation script templates

sadov:
The eden+ports.sh and eden-ports.sh scripts are configured for the EDEN placement.

### DIFF
--- a/tests/eclient/Makefile
+++ b/tests/eclient/Makefile
@@ -49,6 +49,8 @@ test:
 build: setup
 
 setup:
+	$(LOCALBIN) utils template eden+ports.sh.tmpl > eden+ports.sh
+	$(LOCALBIN) utils template eden-ports.sh.tmpl > eden-ports.sh
 	cp eden+ports.sh eden-ports.sh $(BINDIR)/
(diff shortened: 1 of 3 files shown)


tests/eclient/Makefile
```diff
@@ -49,6 +49,8 @@ test:
 build: setup
 
 setup:
+	$(LOCALBIN) utils template eden+ports.sh.tmpl > eden+ports.sh
```
rvs, line 52: Hey @sadov @giggsoff @mydatascience -- can we please add an explicit note about `eden utils template` to our docs some place? It seems super useful and I actually didn't know it existed until now.

rvs (approved):
LGTM

tests/eclient/Makefile
sadov: Actually we have some description about `eden utils template` command and EDEN-specific Go template extension in [tests/README.md](https://github.com/lf-edge/eden/blob/master/tests/README.md#go-template-support), but yes -- may be it would be reasonable make it more informative and visible.

PR #386, opened 2020-11-23
Adding FIO auto-tests

vk-en:
Adding automated performance testing with the FIO utility and publish the results to GitHub.


tests/io_perfomance/image-volume/README.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+> If you want to start a container with the **VOLUME** parameter, you need to specify the tag **volume** in the container. For example: docker://itmoeve/fio_tests:volume
+
+3 required parameters must be specified in the --metadata parameter:
+
+1. **GIT_REPO** - is the name of the repository without .git. For example "eve-performance".
+2. **GIT_LOGIN** - username on GitHub, where the repository specified in GIT_REPO is located
+3. **GIT_TOKEN** - GitHub token for authorization and adding a branch with results to your repository.
+4. **EVE_VERSION** - EVE version. This parameter is required for naming a branch in GitHub.
```
mydatascience, line 22: you can get it. `eden eve version` or even better {{EdenConfig "eve.tag"}}
Let's not ask for something we can find out ourselves =)

tests/io_perfomance/image/config.fio
```diff
@@ -0,0 +1,183 @@
+[global]
```
mydatascience, line 1: Do we really need 2 configs which look the same. 
And two images?

mydatascience (changes requested):

tests/io_perfomance/image-volume/README.md
```diff
@@ -0,0 +1,46 @@
...
+This container runs 36 FIO tests of different types and loads in turn. You can find out more about the test configuration by looking at the config.fio file in the container itself, or the test results in the specified GitHub repository in the directory. Each test is 2 minutes long. The total approximate testing time is ~ 1 hour 12 minutes. Run it and just wait for the result in the specified repository.
+
+## How to
+
+Before starting the container, you need to [get a token on GitHub.com](https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/creating-a-personal-access-token).
+
+### To deploy
+
+```console
+./eden pod deploy --metadata="EVE_VERSION=5.14\nGIT_REPO=eve-performance\nGIT_LOGIN=itmo-eve\nGIT_TOKEN=1111111111111111111111111aaaaaaaaaaaaaaa" -p 8029:80 docker://itmoeve/fio_tests --no-hyper
```
rvs, line 12: As @mydatascience pointed out -- you can discover EVE_VERSION setting.

In fact, the approach I'd recommend @vk-en is to discover/default as much as you can so that I can run this test with only GIT_TOKEN specified and everything else being auto-discovered/defaulted.

tests/io_perfomance/image-volume/README.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+> If you want to start a container with the **VOLUME** parameter, you need to specify the tag **volume** in the container. For example: docker://itmoeve/fio_tests:volume
```
rvs, line 15: This line is unclear to me -- it needs to be elaborated.

tests/io_perfomance/image-volume/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+```console
+./eden test ./tests/io_perfomance
+```
+
+>Before running the test, you need to add environmental variables: GIT_REPO, GIT_LOGIN, GIT_TOKEN.
+
+## About results
+
+At the moment, the test results will be posted to the GitHub repository in a new branch and will have the following tree:
```
rvs, line 36: This description is unclear to me - what does it mean to be posted in a branch? How is the branch name determined? What happens to the other commits that maybe sitting in that same branch?

tests/io_perfomance/image-volume/config.fio
```diff
@@ -0,0 +1,182 @@
...
+iodepth=32
+numjobs=8
+stonewall
+
+[write-1m-4]
+rw=write
+bs=1m
+iodepth=32
+numjobs=1
+stonewall
```
rvs, line 182: While a lot of these settings look reasonable, I wonder where did they come from. Lets discuss how did you decide on picking these very settings and not something else.

tests/io_perfomance/image-volume/run-iostat.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+#Running IOSTAT
+for (( i = 1; i <= 36; i++ ))
+do
+iostat -xm 3 20 > ~/"$GIT_REPO"/"$FOLDERNAME"/Configs/Test-results/Iostat/iostat-fio-group-"$i"
```
rvs, line 6: iostat for now is fine @vk-en but in the long run we will probably need to develop side car performance/observability containers for this kind of a task.

tests/io_perfomance/image-volume/run-iostat.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+#Running IOSTAT
+for (( i = 1; i <= 36; i++ ))
```
rvs, line 4: how is this 1 - 36 determined? I thought you wanted to run run iostat for as long as fio is running, no?

tests/io_perfomance/image-volume/run-test.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
rvs, line 1: I highly recommend you check all your shell code with yetus/shellcheck @vk-en  . Is this on by default in Eden repo @mydatascience ?

tests/io_perfomance/image/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ubuntu:focal
```
rvs, line 1: How big is the resulting container?

tests/io_perfomance/image/README.md
```diff
@@ -0,0 +1,46 @@
+# FIO autotests container
```
rvs, line 1: This looks like exactly the same README as the one in image-volume. Can we please kill one of those?

tests/io_perfomance/image/config.fio
rvs: Same question here -- and see the question about README above.

rvs (changes requested):
This is moving along really nicely -- but still a bit of clean up that needs to be done before this can be merged.

tests/io_perfomance/image-volume/run-iostat.sh
vk-en: We have 36 fio tests in total, so we create separate statistics for each Fio test, collected from Iostat. This way we get more accurate performance data for each Fio test, and not for one file, in which it is difficult to understand in which test we got this or that result.
If the number of tests increases, this range will need to be expanded.

tests/io_perfomance/image/Dockerfile
vk-en: 250mb

tests/io_perfomance/image-volume/README.md
vk-en: When running the test via Eden, we do not require this parameter. I wrote above about the necessary parameters to run the test through eden.
I wrote about it here because I thought it important to tell you what parameters are needed to run the test. As for the example. I'll correct it to {{$ env: = EdenGetEnv "EVE_VERSION"}} or {{EdenConfig "eve.tag"}}.

tests/io_perfomance/image-volume/README.md
vk-en: I will change the parameter for EVE_VERSION in the example.

I would not want to limit this test to one account and a repository, my solution allows you to use any repository and an account with access to it as a target for saving the results. I believe that this will allow this test to pass not only for our employees who have access to the itmo repository

mydatascience: We can't discover repo without GIT_LOGIN / GIT_REPO

PR #397, opened 2020-11-27
Add Documentation about tests in Eden

vk-en:
Add Documentation about tests in Eden


vk-en:
Is it worth adding an example for creating a test (with a simple task)?

mydatascience:
#398

mydatascience:
> Is it worth adding an example for creating a test (with a simple task)?

yess...

PR #422, opened 2020-12-09
[WIP] Add dock for vHost

vk-en:
vHost configuration documentation in EVE

### DIFF
--- a/docs/vhost.md
+++ b/docs/vhost.md
@@ -1,6 +1,6 @@
 # LIO in eve
 
-Инструкция по ручной натсройке vHost в EVE на примере target fileio
+Instruction for manually configuring vHost in EVE using target fileio as an example
 
 ## Example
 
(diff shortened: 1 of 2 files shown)


rvs (changes requested):
Great first cut at the doc. Two tiny nits before this can be merged:
  * please take care of Yetus suggestions
  * can you please rename it to VHOST.md
?

vk-en:
I renamed the file and added more information.

rvs (changes requested):
Why is this coming to Eden instead of EVE?

vk-en:
This guide is an intermediate result of work to understand the configuration vhost, and we thought that it makes no sense to add instructions for manually configuring vHost to the eve repository, because there it will and should be configured automatically.

I also think that while we have not yet decided on the use of vHost, it makes little sense to add manual configuration instructions to the main repository.

But if you think that it is necessary in eve too, then it won't be difficult for me to add it to EVE.

mydatascience:
We have that in eve

mydatascience:
We have

PR #427, opened 2020-12-10
fix quickstart

mydatascience:
Fix Quickstart to run small workload and do not  run setup-onboard tests after quickstart commands


rvs (commented):
LGTM

PR #451, opened 2020-12-22
Added test on GCP & gcp workflow

mydatascience:
Using github Actions to launch EVE on GCP for testing VMs 

One Needs GITHUB secrets 
GCP_PROJECT_ID - google project ID 
GCP_SA_KEY - google project json key 
OVPN_FILE - openvpn file 

Also added new "gcp" workflow as large workflow is too large 

### DIFF
--- a/.github/workflows/eden_gcp.yml
+++ b/.github/workflows/eden_gcp.yml
@@ -10,6 +10,9 @@ jobs:
     steps:
       - name: get eden
         uses: actions/checkout@v2
+      - name: Public IP
+        id: ip
+        uses: haythem/public-ip@v1.2
       - name: setup packages
         run: |
           sudo apt update
@@ -23,9 +26,6 @@ jobs:
           project_id: ${{ secrets.GCP_PROJECT_ID }}
           service_account_key: ${{ secrets.GCP_SA_KEY }}
           export_default_credentials: true
-      - name: Public IP
-        id: ip
-        uses: haythem/public-ip@v1.2
       - name: set firewall & clean
         run: |
           gcloud compute instances delete eve-eden-actions-${{github.run_number}} -q --zone=us-west1-a || echo "not exists"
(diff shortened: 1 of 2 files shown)


rvs:
This is a great start @mydatascience -- a couple of points:
  * please provide a full, detailed list of secrets that this workflow requires and put it in the comment section of the workflow
  * while starting with Eden is great -- we will need that same workflow to be available to us on the EVE side *in addition* to the Eden workflow that we already have
  * for both we should probably transition to using as much cached artifacts as possible (for both EVE and Eden) to avoid wasting time rebuilding everything every time

mydatascience:
GCP_PROJECT_ID - google project ID
GCP_SA_KEY - google project json key
OVPN_FILE - openvpn file (base64 encoded)

mydatascience:
Please review @rvs 
I would say - let's rebuild eden for now.(For eden)
Eve is not built, we use eve tag that is in eden here. 

Once we have it in eden I will make the similar PR to Eve

.github/workflows/eden_gcp.yml
```diff
@@ -0,0 +1,78 @@
...
+          echo "$OVPN_FILE" | base64 -d > ./config.ovpn
+        env:
+          OVPN_FILE: ${{ secrets.OVPN_FILE }}
+      - name: Set up Cloud SDK
+        uses: google-github-actions/setup-gcloud@master
+        with:
+          project_id: ${{ secrets.GCP_PROJECT_ID }}
+          service_account_key: ${{ secrets.GCP_SA_KEY }}
+          export_default_credentials: true
+      - name: Public IP
```
rvs, line 26: Can you please move this step to be right before we divulge our first secret? (setup packages)

rvs (changes requested):

rvs:
Aside from a small change that needs to be implemented -- this looks OK @mydatascience

.github/workflows/eden_gcp.yml
mydatascience: Fixed

rvs:
LGTM. Please let me know what VPN secret should I add now that it has been merged @mydatascience

PR #474, opened 2021-01-12
Main workflow and tests extensions and fixes

sadov:
Added USB Passthtrough test and some fixes related to `eve.devmodel` checking. [Eclients](https://github.com/itmo-eve/eden/blob/tests_extension/tests/eclient/testdata/eclients.txt) stress test will run now only with non zero EDEN_TEST_APPS env. var.

Added tests:
* [usb-pt_test](https://github.com/itmo-eve/eden/blob/tests_extension/tests/eclient/testdata/usb-pt_test.txt) -- USB Passthtrough test;
* [eve_restart](https://github.com/itmo-eve/eden/blob/tests_extension/tests/workflow/testdata/eve_restart.txt) -- Eve restart after config changes for QEMU variant.

### DIFF
--- a/pkg/models/common.go
+++ b/pkg/models/common.go
@@ -115,17 +115,21 @@ func generatePhysicalIOs(ethCount, wifiCount, usbCount uint) []*config.PhysicalI
 			},
 		})
 	}
+	usbGroup := 0
 	for i := uint(0); i < usbCount; i++ {
-		num := fmt.Sprintf("%d:1", i)
-		name := fmt.Sprintf("USB%s", num)
-		physicalIOs = append(physicalIOs, &config.PhysicalIO{
-			Ptype:        evecommon.PhyIoType_PhyIoUSB,
-			Phylabel:     name,
-			Logicallabel: name,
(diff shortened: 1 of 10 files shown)


rvs (changes requested):
The test itself is OK, but I'm concerned that we're now bifurcating the tests into being dependent on what environment they get to be executed in. This one, for example, clearly requires qemu. Before this gets merged I think we need to have a conversation on how we plan to manage these kinds of dependencies.

sadov:
Sure, it seems ugly for me too. Generally, this check can be moved inside the test. But we need to do some preliminary configuration for QEMU too and restart it. Better to do it once, and now it happens in the overall workflow script. Maybe someone has an idea how best to handle different flavors of EVE platforms in this case.

mydatascience:
Maybe use just workflow - qemu and this test will work only for this workflow. Rely on devmodel is a bit complicated and definitely need some explanation...

mydatascience:
@sadov  Need to fix xen on gcp

Issue #126, opened 2020-06-29
Actual status of EVE if it's unavalable in eden status


mydatascience:
Once onboarded, remote EVE doesn't change the status even if it has gone offline.
We need to find some discovery mechanism to get the actual  remote EVE status and display it in `eden status`

rvs:
Hey @mydatascience -- can you guys prioritize this a bit? This is especially relevant given recent confusion about the need for `eden eve onboard` command. So my suggestion would be to include that help text into eden status output:

```
eden status
...
EVE is running ... (but it isn't onboarded -- make sure to run  eden eve onboard to complete the onboarding sequence)
```

PR #521, opened 2021-02-13
Assume EVE is down if no answer comes in 10 min

mydatascience:
Final fix for https://github.com/lf-edge/eden/issues/126

Assume that EVE is down if in 10 min no answer comes (Last seen is too late)
We may change it to 5 min 

@giggsoff 


giggsoff (approved):
Looks reasonable.

PR #529, opened 2021-02-19
[WIP] Add initial maintenance mode test

eriknordmark:
Have issues that the status from the device does not appear in the test.

The idea is to evolve the test (once it can toggle maintenance_mode) to check that config changes will be ignored  by the device but that the device can be rebooted while it is in maintenance mode.


mydatascience:
Should we do something with this PR? @eriknordmark

eriknordmark:
Closing as abandoned — this WIP from 2022 has been superseded by other test work.

PR #531, opened 2021-02-22
add registry option to controller

deitch:
When passing an OCI url, e.g. `oci://docker.io/foo/bar:1.2.3`, to launching pods, we already have an option to pass it `--registry=local` so that it uses the local registry. This has been useful for images that are built locally and not available on the remote registry hub.

This adds the same `--registry` option to `eveimage-update`, so we can do the same.


deitch:
Just tested, does the right thing. Actually, it was really nice to be able to load an image from local docker to local eden registry and then tell eve to pull it for baseos.

PR #576, opened 2021-04-02
Port forwarding test between apps connected to different network instances

gkodali-zededa:
1) Make both Qemu and VBox connect to same network from two different ports.
2) Enhance port forwarding test to try connecting across network instances.

Needs to be seen how parallels can be made to do the same.


giggsoff (changes requested):
Unfortunately, it does not work for me, probably due to hardcoded overlapping subnets on interfaces.

```
        > exec -t 1m bash test_connectivity.sh 2234 2223
        [stdout]
        192.168.0.10
        ssh -o ConnectTimeout=10 -o StrictHostKeyChecking=no -i /Users/petrfedchenkov/GolandProjects/eden/dist/tests/eclient/image/cert/id_rsa root@127.0.0.1 -p 2234 sh /root/portmap_test.sh 2223
        ssh -o ConnectTimeout=10 -o StrictHostKeyChecking=no -i /root/.ssh/id_rsa root@192.168.0.10 -p 2223 grep Ubuntu /etc/issue
        [stderr]
        @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
        @    WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!     @
        @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
        IT IS POSSIBLE THAT SOMEONE IS DOING SOMETHING NASTY!
        Someone could be eavesdropping on you right now (man-in-the-middle attack)!
        It is also possible that a host key has just been changed.
        The fingerprint for the ECDSA key sent by the remote host is
        SHA256:hYSnInDfR3z6drvLPzQCcF3Zq3Q1d75Z73ZCCN5SWrs.
        Please contact your system administrator.
        Add correct host key in /Users/petrfedchenkov/.ssh/known_hosts to get rid of this message.
        Offending ECDSA key in /Users/petrfedchenkov/.ssh/known_hosts:4
        Password authentication is disabled to avoid man-in-the-middle attacks.
        Keyboard-interactive authentication is disabled to avoid man-in-the-middle attacks.
        ssh: connect to host 192.168.0.10 port 2223: Connection timed out
        [exit status 255]
        FAIL: ../eclient/testdata/port_forward.txt:87: command failure
```

I pushed new version of eclient, can you please also try?

pkg/eden/eden.go
```diff
@@ -364,13 +364,52 @@ func StartEVEVBox(vmName, eveImageFile string, cpus int, mem int, hostFwd map[st
 			log.Fatalf("VBoxManage error for command %s %s", commandArgsString, err)
 		}
 
-		commandArgsString = fmt.Sprintf("modifyvm %s  --nic1 nat --cableconnected1 on", vmName)
+		commandArgsString = fmt.Sprintf("natnetwork add --netname %s --network %s --enable --dhcp on",
+										"natnet1", "10.0.2.0/24")
```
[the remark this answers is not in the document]
gkodali-zededa: Both NICs use the same subnet. This is the same subnet that vboxmanage is hardcoded to create. We have to provide a subnet to create network.

tests/eclient/testdata/port_forward.txt
```diff
@@ -76,12 +120,14 @@ do
...
+-- test_connectivity.sh --
 EDEN={{EdenConfig "eden.root"}}/{{EdenConfig "eden.bin-dist"}}/{{EdenConfig "eden.eden-bin"}}
 HOST=$($EDEN eve ip)
+EVE_IP=$({{template "ssh"}}$HOST -p $2 sh /root/get_eve_ip.sh)
+echo "$EVE_IP"
 
-echo {{template "ssh"}}$HOST -p $1 sh /root/portmap_test.sh -p $2
-{{template "ssh"}}$HOST -p $1 sh /root/portmap_test.sh $2
+echo {{template "ssh"}}$HOST -p $1 sh /root/portmap_test.sh $2
+{{template "ssh"}}$HOST -p $1 sh /root/portmap_test.sh $EVE_IP $2
```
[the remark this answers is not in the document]
gkodali-zededa: will fix

pkg/eden/eden.go
```diff
@@ -436,19 +527,47 @@ func StartEVEQemu(qemuARCH, qemuOS, eveImageFile, qemuSMBIOSSerial string, eveTe
...
-	for ind, n := range nets {
-		qemuOptions += fmt.Sprintf("-netdev user,id=eth%d,net=%s,dhcpstart=%s", ind, n.Subnet, n.FirstAddress)
-		if ind == 0 {
-			for k, v := range qemuHostFwd {
-				qemuOptions += fmt.Sprintf(",hostfwd=tcp::%s-:%s", k, v)
-			}
+	offset := 0
+	firstNet := nets[0].Subnet
+	firstAddr := nets[0].FirstAddress
+	secondAddr := net.ParseIP("192.168.0.11")
```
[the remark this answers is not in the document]
gkodali-zededa: Existing logic created two different subnets and packets do not get routed between these subnets. Intention is to create a single subnet and keep both interfaces in that subnet. I did not understand the novel offsets part. Can you please elaborate?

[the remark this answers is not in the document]
giggsoff: I mean, that before your changes we had nics on different subnets provided by `utils.GetSubnetsNotUsed` function. With your change we have one subnet (`firstNet`) which depends on environment (192.168.0.1/24 in my case) and another one is hardcoded (`net.ParseIP("192.168.0.11")`). Are we really expecting this? What is the point of putting a harcode in this line?
> I did not understand the novel offsets part. Can you please elaborate?

If I correctly understood, your add port forwarding on the second nic with predefined offset from ports in the config. In this case why not to add this logic into existing code
https://github.com/lf-edge/eden/blob/125938afbd50da1299d6ecdc97054b9f2e04353c/pkg/eden/eden.go#L447

>Intention is to create a single subnet and keep both interfaces in that subnet.

Could you elaborate why this is necessary?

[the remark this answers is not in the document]
gkodali-zededa: utils.GetSubnetsNotUsed creates subnet and that is 192.169.0.0/24 and it also allocates an IP i.e. 192.168.0.10 for the first NIC. The 192.168.0.11 that is see is not a subnet, but an IP used for the second NIC.
Both NICs have to be in the same subnet, because if they are in different subnets, qemu networking is not routing packets between them. Instead packets are going somewhere out into the internet. Probably that is because of the separate NAT instances created for these subnets. Also port forwarding has the case where port forwarded packets pass between two different interfaces in the same subnet. Basically apps connected to different network instances, but with uplink ports connected to the same network should be able to talk to each other using port forwarding.

gkodali-zededa:
> Unfortunately, it does not work for me, probably due to hardcoded overlapping subnets on interfaces.
> 
> ```
>         > exec -t 1m bash test_connectivity.sh 2234 2223
>         [stdout]
>         192.168.0.10
>         ssh -o ConnectTimeout=10 -o StrictHostKeyChecking=no -i /Users/petrfedchenkov/GolandProjects/eden/dist/tests/eclient/image/cert/id_rsa root@127.0.0.1 -p 2234 sh /root/portmap_test.sh 2223
>         ssh -o ConnectTimeout=10 -o StrictHostKeyChecking=no -i /root/.ssh/id_rsa root@192.168.0.10 -p 2223 grep Ubuntu /etc/issue
>         [stderr]
>         @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
>         @    WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!     @
>         @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
>         IT IS POSSIBLE THAT SOMEONE IS DOING SOMETHING NASTY!
>         Someone could be eavesdropping on you right now (man-in-the-middle attack)!
>         It is also possible that a host key has just been changed.
>         The fingerprint for the ECDSA key sent by the remote host is
>         SHA256:hYSnInDfR3z6drvLPzQCcF3Zq3Q1d75Z73ZCCN5SWrs.
>         Please contact your system administrator.
>         Add correct host key in /Users/petrfedchenkov/.ssh/known_hosts to get rid of this message.
>         Offending ECDSA key in /Users/petrfedchenkov/.ssh/known_hosts:4
>         Password authentication is disabled to avoid man-in-the-middle attacks.
>         Keyboard-interactive authentication is disabled to avoid man-in-the-middle attacks.
>         ssh: connect to host 192.168.0.10 port 2223: Connection timed out
>         [exit status 255]
>         FAIL: ../eclient/testdata/port_forward.txt:87: command failure
> ```
> 
> I pushed new version of eclient, can you please also try?

It's a bit odd. The same test with qemu passed on Linux and fails on Mac. 
The Qemu user network on linux switches packets between hosts connected to the same network. The same does not happen with Qemu on mac. On mac packets are instead getting routed out of the host machine. Hence the test fails. Need to dig more into this.

gkodali-zededa:
> Unfortunately, it does not work for me, probably due to hardcoded overlapping subnets on interfaces.
> 
> ```
>         > exec -t 1m bash test_connectivity.sh 2234 2223
>         [stdout]
>         192.168.0.10
>         ssh -o ConnectTimeout=10 -o StrictHostKeyChecking=no -i /Users/petrfedchenkov/GolandProjects/eden/dist/tests/eclient/image/cert/id_rsa root@127.0.0.1 -p 2234 sh /root/portmap_test.sh 2223
>         ssh -o ConnectTimeout=10 -o StrictHostKeyChecking=no -i /root/.ssh/id_rsa root@192.168.0.10 -p 2223 grep Ubuntu /etc/issue
>         [stderr]
>         @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
>         @    WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!     @
>         @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
>         IT IS POSSIBLE THAT SOMEONE IS DOING SOMETHING NASTY!
>         Someone could be eavesdropping on you right now (man-in-the-middle attack)!
>         It is also possible that a host key has just been changed.
>         The fingerprint for the ECDSA key sent by the remote host is
>         SHA256:hYSnInDfR3z6drvLPzQCcF3Zq3Q1d75Z73ZCCN5SWrs.
>         Please contact your system administrator.
>         Add correct host key in /Users/petrfedchenkov/.ssh/known_hosts to get rid of this message.
>         Offending ECDSA key in /Users/petrfedchenkov/.ssh/known_hosts:4
>         Password authentication is disabled to avoid man-in-the-middle attacks.
>         Keyboard-interactive authentication is disabled to avoid man-in-the-middle attacks.
>         ssh: connect to host 192.168.0.10 port 2223: Connection timed out
>         [exit status 255]
>         FAIL: ../eclient/testdata/port_forward.txt:87: command failure
> ```
> 
> I pushed new version of eclient, can you please also try?

For some reason networking with Qemu on mac does not work the same as Qemu on Linux. For now, the new test has been restricted to run only if the host OS is LInux.

pkg/eden/eden.go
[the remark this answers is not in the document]
gkodali-zededa: Will also make this a constant in pkg/defaults package.

pkg/expect/expectation.go
```diff
@@ -131,15 +126,6 @@ func AppExpectationFromURL(ctrl controller.Cloud, device *device.Ctx, appLink st
...
 			}
-			if len(qemuPorts) > 0 { //not empty forwarding rules, need to check for existing
-				for _, qv := range qemuPorts {
-					if qv == strconv.Itoa(extPort) {
-						ni.ports[extPort] = intPort
-						continue exit
-					}
-				}
-				log.Fatalf("Cannot use external port %d. Not in Qemu %s", extPort, qemuPorts)
-			}
```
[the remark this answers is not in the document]
gkodali-zededa: will do

cmd/edenStart.go
```diff
@@ -122,6 +124,30 @@ var startCmd = &cobra.Command{
 	},
 }
 
+func getUplinkPortIPMap() map[string]net.IP {
```
mydatascience, line 127: Does this function belong here? 
It looks it should be wrapped somewhere around set-ports.sh ?

pkg/eden/eden.go
```diff
@@ -341,7 +341,7 @@ func StatusEServer() (status string, err error) {
 }
```
mydatascience, line 341: Same large comment - let's separate the functions and implementation-specific things. 
I'd like to see 
  Vbox.go
  Parallels.go
  Qemu.go

pkg/utils/networking.go
```diff
@@ -32,15 +33,21 @@ func getSubnetByInd(ind int) (*net.IPNet, error) {
...
 	if ind < 0 || ind > 255 {
 		return nil, fmt.Errorf("error in index %d", ind)
 	}
 	IP := net.ParseIP(fmt.Sprintf("192.168.%d.10", ind))
 	if IP == nil {
 		return nil, fmt.Errorf("error in ParseIP for index %d", ind)
 	}
-	return IP, nil
+	ips := []net.IP{IP}
+	IP2 := net.ParseIP(fmt.Sprintf("192.168.%d.11", ind))
```
mydatascience, line 45: Magic number?

tests/eclient/image/get_eve_ip.sh
```diff
@@ -0,0 +1,4 @@
+#! /bin/sh
+
+EVE_IP=$(curl -s "http://169.254.169.254/eve/v1/network.json" | jq -r '."external-ipv4"')
```
mydatascience, line 3: Looks it's a strange magic number. Maybe even wrong?

mydatascience (changes requested):

cmd/edenStart.go
gkodali-zededa: I am lost. Please help me here.
This function makes API call to controller for the lastest info message that the device sent.
We read the IP addresses that the edge-node got from DHCP from this reponse. These IP addresses are used for setting up port map rules during the test.
Should this be done using shell script and pass the information to go program somehow?

pkg/expect/expectation.go
```diff
@@ -133,12 +133,15 @@ func AppExpectationFromURL(ctrl controller.Cloud, device *device.Ctx, appLink st
 			}
 			if len(qemuPorts) > 0 { //not empty forwarding rules, need to check for existing
 				for _, qv := range qemuPorts {
-					if qv == strconv.Itoa(extPort) {
+					portNum, err := strconv.Atoi(qv)
+					if err != nil {
+						log.Fatalf("Port map port %s could not be converted to Integer", qv)
+					}
+					if portNum == extPort || (portNum + defaults.DefaultPortMapOffset) == extPort {
```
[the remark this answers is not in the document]
gkodali-zededa: I'll see if I can avoid this.

pkg/utils/networking.go
gkodali-zededa: No. It is an IP address. Needed to allocate another static IP address for a different ethernet interface on the emulated machine. Copied from what is there 4 lines above.

tests/eclient/image/get_eve_ip.sh
gkodali-zededa: https://serverfault.com/questions/427018/what-is-this-ip-address-169-254-169-254

gkodali-zededa: EVE is the meta-data distributor in this case. Please let me know if it still looks wrong.

eriknordmark: @mydatascience The eve-specific URL is documented in https://github.com/lf-edge/eve/blob/master/docs/ECO-METADATA.md The generic use of http://169.254.169.254/ is part of cloud-init.

pkg/eden/eden.go
gkodali-zededa: will do that

gkodali-zededa: made separate files

tests/eclient/image/get_eve_ip.sh
mydatascience: got it

pkg/utils/networking.go
mydatascience: got it.

mydatascience:
Please resolve the conflicts and let's merge. But I am still confused with the calculation for the shifts + ports.

gkodali-zededa:
> Please resolve the conflicts and let's merge. But I am still confused with the calculation for the shifts + ports.

I resolved the conflicts and the PR build is currently running.

The port maps list is right now considered global and the same port maps are added on all ethernet interfaces of the emulated machine. With this behavior there will be port map conflicts. For my test I didn't want that to happen and so decided to add an offset for each interface (keeping the port map list unchanged).
At the same time I understand your concern that it is a bit confusing. It needs to be made better, but should be taken as a separate task to make a port map list per emulated (edge-node) interface.

pkg/eden/vbox.go
```diff
@@ -0,0 +1,251 @@
...
+			}
+
+			commandArgsString = fmt.Sprintf("modifyvm %s --nic2 natnetwork --nat-network2 %s --cableconnected2 on",
+			vmName, "natnet1")
+			if err = utils.RunCommandWithLogAndWait("VBoxManage", defaults.DefaultLogLevelToPrint, strings.Fields(commandArgsString)...); err != nil {
+				log.Fatalf("VBoxManage error for command %s %s", commandArgsString, err)
+			}
+		}
+
+		eth0IP := ipMap["eth0"]
```
giggsoff, line 147: Can we have fallback variant here(with empty ipMap)? In your variant, we expect ipMap filled, but it is true only after EVE onboard.

pkg/eden/vbox.go
```diff
@@ -0,0 +1,251 @@
...
+				log.Errorf("Parsing %s to Integer value failed", k)
+				break
+			}
+			hostPort += 10
+			guestPort, err := strconv.Atoi(v)
+			if err != nil {
+				log.Errorf("Parsing %s to Integer value failed", v)
+				break
+			}
+			guestPort += 10
```
giggsoff, line 84: We have default for this. And the same point for another parts.

giggsoff (changes requested):
Please take a look onto ipMap for VirtualBox, seems, we will have no port map without restart of EVE after onboard.

pkg/eden/vbox.go
gkodali-zededa: It's not clear to me. Can you please elaborate?

giggsoff: You add `getUplinkPortIPMap()` function, which will return non-empty map only in case of onboarded EVE. But in the beginning (not onboarded, just started) it will be empty. So, we should care about it.

gkodali-zededa: Yes, you are right. We'll need reboot for the port maps to take affect. Before device onboard, the rules will have the IP as empty string. This should not adversely affect us because we do not run tests before the device in onboarded. After the device is onboarded and rebooted, the port map rules will have valid IP addresses. We already have a reboot in the beginning of tests/workflow that helps setup the port maps properly in this case.
With virtual box, unfortunately we cannot just tell the destination port for packet while configuring host port map rules. It will require the destination IP address for configuring port maps. It's easier with qemu but a bit of inconvenient with vbox

giggsoff: Sure, but Eden is not only about the workflow.  Ok, if you think, it is difficult to workaround empty ipMap here, we can think about it in next PRs.

gkodali-zededa: My point was that it did not make any difference if we handled the empty ip map or not. It would still require a reboot before the port maps can take effect. Do you have something that proves contrary?

pkg/eden/vbox.go
gkodali-zededa: The discussion above applies to this also.

tests/eclient/testdata/port_forward.txt
```diff
@@ -17,33 +17,80 @@ message 'Resetting of EVE'
...
+! stdout '^n1\s'
+
+############################## TEST SCENARIO 2 ################################
+# Hairpin connectivity test between two apps connected to two different network instances.
+# These two network instances have different uplink adapters attached to them. And these adapters
+# get different IP addresses from the same subnet with DHCP.
+
+{{$osruntime := EdenOSRuntime "nan"}}
+
+{{if (eq $osruntime "linux")}}
```
giggsoff, line 68: If I correctly understood, we rely here on network environment generated with qemu/VirualBox. But we have tests on GCP for now.
Can you please add here checks similar with checks in the workflow https://github.com/lf-edge/eden/blob/f2c9e5ed8937f80499e05afb8a08df44323c404d/tests/workflow/eden.workflow.tests.txt#L41

dautovri:
@gkodali-zededa Please consider this ticket https://www.virtualbox.org/ticket/20175. We still have problems running on macOS. Fix will reach the release version soon, but not there yet. 
image

tests/eclient/testdata/port_forward.txt
gkodali-zededa: Done. Pushed the changes.

gkodali-zededa:
> @gkodali-zededa Please consider this ticket https://www.virtualbox.org/ticket/20175. We still have problems running on macOS with VirtualBox. Fix will reach the release version soon, but not there yet.
> 
> image

Petr gave me the URL for fixed build and I currently use that build on mac. Thanks for letting me know.

giggsoff (approved):
Thanks for the edits and your patience!

gkodali-zededa:
> Thanks for the edits and your patience!

Can we merge this now?

PR #606, opened 2021-04-30
[WIP] eth reboot test

dautovri:
Assumes a second connected Ethernet. Key purpose is to verify that a reboot (sudo shutdown -r +1) from the guest doesn’t disrupt the set of assigned adapters. (We had bugs there in the past.) This test verifies that the application can communicate over the assigned Ethernet, and that it can not communicate via the NAT in EVE (due to the ACL blocking communication via the default network instance.) Note that the deployed application has two Ethernet interfaces.

Signed-off-by: Ruslan Dautov <dautov2@gmail.com>

### DIFF
--- a/tests/eclient/qemu+eth.sh
+++ b/tests/eclient/qemu+eth.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+
+test -n "$EDEN_CONFIG" || EDEN_CONFIG=default
+
+EDEN=eden
+DIR=$(dirname "$0")
+PATH=$DIR:$DIR/../../bin:$PATH
+
+dist=$($EDEN config get "$EDEN_CONFIG" --key eden.root)
+
+dd if=/dev/zero of="$dist/stick.raw" bs=1K count=1
+
+cat >> ~/.eden/"$EDEN_CONFIG"-qemu.conf <<END
+
+
+[netdev "hostnet"]
+  type = "user"
+
+[device "net"]
+  driver = "virtio-net-pci"
+  netdev = "hostnet"
+  bus = "pcie.0"
+  addr = "19.0"
+END
+
+cat <<END
+To activate the changes in the config, you need to restart EVE:
+  $EDEN eve stop
+  $EDEN eve start
+END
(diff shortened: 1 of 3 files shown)


dautovri:
Every uplink ethX interface is renamed to kethX and put under a new bridge created with the original name ethX.
IP and mac addresses are moved from keth to eth (from uplink interface to bridge).
So when a  network of type "switch" is created, it has the bridge linked with the upstream interface already prepared. 

With this PR via ./qemu+eth.sh we creating a new eth interface in `default-qemu.conf` by adding 
```
[netdev "hostnet"]
  type = "user"

[device "net"]
  driver = "virtio-net-pci"
  netdev = "hostnet"
  bus = "pcie.0"
  addr = "19.0"
```

At the moment I discover that keht2 sometimes is not creating.

tests/eclient/qemu+eth.sh
```diff
@@ -0,0 +1,30 @@
...
+  netdev = "hostnet"
+  bus = "pcie.0"
+  addr = "19.0"
+END
+
+cat <<END
+To activate the changes in the config, you need to restart EVE:
+  $EDEN eve stop
+  $EDEN eve start
+END
```
milan-zededa, line 30: As an alternative to this qemu-specific script, you could instead try to reuse already existing `eth1` interface for this test. The challenge is that all device models that have `eth1` (parallels, qemu, vbox), configure `eth1` as a management uplink (https://github.com/lf-edge/eden/blob/master/pkg/models/common.go#L86).
For your test you could try to:
- prepare custom device model (as json file referenced by `eve.devmodelfile`) - it would be pretty much the same as already used for qemu, vbox and parallels, with only `usage` of `eth1` modified to `PhyIoUsageNone`
- inside the test, set eden to use the custom model and reset eve (https://github.com/lf-edge/eden/blob/master/docs/eve-models.md#file-to-overwrite-model-settings)
- in the end of the test revert back to the original device model

Not sure if feasible and how much more complicated it would be than the current solution, just worth to consider/try.
As an added bonus the `eve.devmodelfile` option would become part of the testing.

PR #611, opened 2021-05-05
Add ethernet default cost parameter to models

gkodali-zededa:
Signed-off-by: gkodali-zededa <gkodali@zededa.com>

### DIFF
--- a/models/template_l1_FWA-1012VC.json
+++ b/models/template_l1_FWA-1012VC.json
@@ -118,7 +118,7 @@
         "PciLong": "0000:00:15.0"
       },
       "logicallabel": "wwan0",
-      "cost": 0,
+      "cost": 10,
       "usagePolicy": {}
     },
     {
@@ -130,7 +130,7 @@
         "PciLong": "0000:00:15.0"
       },
       "logicallabel": "wwan1",
-      "cost": 0,
+      "cost": 10,
       "usagePolicy": {}
     },
     {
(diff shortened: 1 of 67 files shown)


models/template_l1_FWA-1012VC.json
```diff
@@ -111,6 +118,7 @@
         "PciLong": "0000:00:15.0"
       },
       "logicallabel": "wwan0",
+      "cost": 0,
```
eriknordmark, line 121: We shouldn't default this to zero for wwan

models/template_l1_SYS-E100-9APP-wwan.json
```diff
@@ -49,6 +51,7 @@
         "PciLong": "0000:00:15.0"
       },
       "logicallabel": "wwan0",
+      "cost": 0,
```
eriknordmark, line 54: Ditto

eriknordmark (changes requested):
Please pick a non-zero number than zero for wwan0/wwan1.
How about using 10 for those?

models/template_l1_SYS-E100-9APP-wwan.json
gkodali-zededa: fixed. Made it 10

models/template_l1_FWA-1012VC.json
gkodali-zededa: fixed. Made it 10

gkodali-zededa:
> Please pick a non-zero number than zero for wwan0/wwan1.
> How about using 10 for those?

Done. Made it 10

eriknordmark (approved):
LGTM

Issue #595, opened 2021-04-23
VirtualBox eve onboard issue 


dautovri:
# Description

Eve problem with onboard. VirtualBox devmodel.

VirtualBox 6.1.20r143896
OS: MacOS BigSur 11.2.3
Device:  Macbook Pro 13 2017 

# Steps to Reproduce
```
git clone https://github.com/lf-edge/eden.git
cd eden
make clean
make build
eden config add default --devmodel VBox
eden setup
eden start
eden eve onboard
```

# Actual Results
```
 ./eden eve onboard
INFO[0000] Adam waiting for EVE registration (0) of (20) 
INFO[0020] Adam waiting for EVE registration (1) of (20) 
INFO[0040] Adam waiting for EVE registration (2) of (20) 
INFO[0060] Adam waiting for EVE registration (3) of (20) 
INFO[0080] Adam waiting for EVE registration (4) of (20) 
INFO[0100] Adam waiting for EVE registration (5) of (20) 
INFO[0120] Adam waiting for EVE registration (6) of (20) 
INFO[0140] Adam waiting for EVE registration (7) of (20) 
INFO[0160] Adam waiting for EVE registration (8) of (20) 
INFO[0180] Adam waiting for EVE registration (9) of (20) 
INFO[0200] Adam waiting for EVE registration (10) of (20) 
INFO[0220] Adam waiting for EVE registration (11) of (20) 
INFO[0240] Adam waiting for EVE registration (12) of (20) 
INFO[0260] Adam waiting for EVE registration (13) of (20) 
INFO[0280] Adam waiting for EVE registration (14) of (20) 
INFO[0300] Adam waiting for EVE registration (15) of (20) 
INFO[0320] Adam waiting for EVE registration (16) of (20) 
INFO[0340] Adam waiting for EVE registration (17) of (20) 
INFO[0360] Adam waiting for EVE registration (18) of (20) 
INFO[0380] Adam waiting for EVE registration (19) of (20) 
```

dautovri:
Fixed. https://github.com/lf-edge/eden/pull/636

PR #636, opened 2021-05-28
Fix port forwarding on VBox

milan-zededa:
This PR attempts to fix the issue with port forwarding on VirtualBox.
For port forwarding rules Vbox requires to provide the guest IP(s). Currently these are obtained from the adam controller. But because during startup EVE is not yet onboarded, eden fails to get guest IPs and as a result installs port forwarding rules with empty guest IPs, which is invalid and does not work. The current fix/workaround is to restart EVE VM - on the second start EVE's IPs are known to the controller (if there is onboarding between the first start and the restart).

With this PR eden obtains EVE uplink IPs from the leases of the VBox DHCP server instead. As a result there is some waiting added during `eve start` for EVE to boot and DHCP to allocate IPs (we need to wait for it anyway for on-boarding to proceed). But restarts are quick, Vbox DHCP server preserves leases even if the network was recreated.

Verified on Linux and will test on MacOS over the weekend.

### DIFF
--- a/cmd/edenStart.go
+++ b/cmd/edenStart.go
@@ -3,8 +3,6 @@ package cmd
 import (
 	"fmt"
 	"github.com/lf-edge/eden/pkg/eden"
-	"github.com/lf-edge/eden/pkg/eve"
-	"net"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -111,7 +109,7 @@ var startCmd = &cobra.Command{
 				log.Infof("EVE is starting in Parallels")
 			}
 		} else if devModel == defaults.DefaultVBoxModel {
-			if err := eden.StartEVEVBox(vmName, eveImageFile, cpus, mem, hostFwd, getUplinkPortIPMap()); err != nil {
+			if err := eden.StartEVEVBox(vmName, eveImageFile, cpus, mem, hostFwd); err != nil {
 				log.Errorf("cannot start eve: %s", err)
 			} else {
 				log.Infof("EVE is starting in Virtual Box")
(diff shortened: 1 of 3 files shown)


pkg/eden/vbox.go
```diff
@@ -123,89 +77,173 @@ func StartEVEVBox(vmName, eveImageFile string, cpus int, mem int, hostFwd map[st
...
+		if ipAddr == "" {
+			err = fmt.Errorf("failed to get IP address of eth%d", i)
+			return
+		}
+	}
+	return
 }
 
 //StopEVEVBox function stop EVE in VirtualBox
 func StopEVEVBox(vmName string) (err error) {
```
dautovri, line 237: We will need to make  `VBoxManage dhcpserver remove ` 
Otherwise, we will get following error when we will start again 
```
VBoxManage: error: Cannot start DHCP server because it is already running (pid 0000070d)
VBoxManage: error: Details: code VBOX_E_OBJECT_IN_USE (0x80bb000c), component DHCPServerWrap, interface IDHCPServer, callee nsISupports
VBoxManage: error: Context: "Start()" at line 482 of file VBoxManageNATNetwork.cpp
VBoxManage: error: Failed to start network
FATA[0013] VBoxManage error for command natnetwork start --netname natnet1 exit status 2 

```

dautovri (commented):
Tested on macOS Big Sur, confirming its works. Yes, it takes a longer time on the first start and onboard, second start is fast. 
Some comments about DHCP server workflow.

pkg/eden/vbox.go
milan-zededa: Thanks @dautovri for the feedback. Added explicit removal of the DHCP server and successfully verified that EVE restarts now works on both Linux and macOS.

dautovri:
Tested latest commits, LGTM. 
We can close https://github.com/lf-edge/eden/issues/595 as well.

milan-zededa:
ACL test is failing - needs to be rebased on top of https://github.com/lf-edge/eden/pull/637

rvs (approved):
LGTM

cmd/edenStart.go
```diff
@@ -111,7 +109,7 @@ var startCmd = &cobra.Command{
 				log.Infof("EVE is starting in Parallels")
 			}
 		} else if devModel == defaults.DefaultVBoxModel {
-			if err := eden.StartEVEVBox(vmName, eveImageFile, cpus, mem, hostFwd, getUplinkPortIPMap()); err != nil {
```
mydatascience, line 114: Doesn't it break something. I think we did this getUplinkPortIPMap for the case when we have several adapters

milan-zededa: We still have several network adapters (two exactly: `eth0` and `eth1`). What changed is how we get DHCP-allocated IP addresses for those adapters. `getUplinkPortIPMap` would get this information from the controller, but this doesn't work for the first `./eden eve start` because device is not yet onboarded at that point. With this PR DHCP information is obtained directly from the Vbox DHCP server instead.

mydatascience: what if it's not Vbox

milan-zededa: For qemu and parallels it is not needed specify guest IP in port forwarding rules.
Do you want me to leave `getUplinkPortIPMap` there even if it is not needed anymore?

mydatascience: I'm ok for that

PR #642, opened 2021-06-04
README: fix example port

JohnDMcMaster:
Several places document how to do something and then follow with an example. However, the port in the documentation and following examples aren't consistent. Looks like convention is port 8027 for SSH and 8028 for HTTP, so I've adjusted them to be consistent with port 8028 as they are showing how to use the HTTP interface in these sections.


rvs:
Thanks for taking time to fix this @JohnDMcMaster ! Much appreciated!

PR #649, opened 2021-06-16
increase sleeps and timeouts in tests for rpi4

Insei:
When testing the EVE releases on Rpi4, i found floating test errors associated with small timeout values and small sleep values ​​in the tests. This PR fixes the problem.

### DIFF
--- a/tests/eclient/testdata/acl.txt
+++ b/tests/eclient/testdata/acl.txt
@@ -22,7 +22,7 @@ exec chmod 600 {{EdenConfig "eden.tests"}}/eclient/image/cert/id_rsa
 
 message 'Resetting of EVE'
 eden eve reset
-exec sleep 20
+exec sleep 30
 
 # use zededa.com IP address as a target for $fake_domain
 exec -t 10m bash dns_lookup.sh zededa.com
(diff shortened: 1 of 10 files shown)


milan-zededa:
@Insei Seems that workflow caught an issue which is not related to you PR - eve failed to recover from a network outage, which is being exercised by `port_forward` test. I will have a look. But also there seems to be a bug in the test framework:
`exec -t 5m bash wait_ssh.sh 2223 2234` should timeout after 5 minutes, but it keeps going until it is killed by github itself.

Insei:
@milan-zededa I mean my local tests more. In which tests fail due to obvious errors. For example ./eden network ls takes 8 seconds, not 5(https://github.com/lf-edge/eden/pull/649/files#diff-d068babca59eae706e6e1f33890c29a9f46cebf6bcebc6fe1e48d189a2c71b9eR1), which results in test errors and the test crashes. Also, rpi4 needs more time after ./eden eve reset, otherwise an error may occur when creating app instances.

mydatascience:
@Insei  please fix integration tests

Insei:
@mydatascience fixed

PR #652, opened 2021-06-30
Test for switch network instance VLANs

milan-zededa:
Test deploys multiple applications into a single switch network sliced using VLANs. It also deploys an app providing DHCP services separately for every VLAN. Then it checks if IP address allocations are as expected and that only applications inside the same VLAN are able to talk to each other.

To run the test standalone (i.e. outside of the workflow), execute:
```
 export EDEN_FAIL_RESET=y
 make clean && make build-tests
 ./eden config add default
 ./eden config set default --key eve.log-level --value debug
 ./eden setup
 ./dist/bin/eden+ports.sh 2223:2223 2224:2224 5912:5902 5911:5901 8027:8027 8028:8028 8029:8029 8030:8030 8031:8031
 ./eden start
 ./eden eve onboard
 ./eden test tests/network -e vlans -v debug
```

Note: test depends on the DHCP-server app being able to create VLAN sub-interfaces, which requires NETADMIN capabilities.
For security reasons, EVE does not allow to grant these capabilities to native containers, and so it is required to deploy the app as a VM-in-Container. Without hardware assisted virtualization the test is therefore automatically skipped (it will not be included in the integration check, but will run after merge on GCP).

### DIFF
--- a/tests/network/testdata/vlans.txt
+++ b/tests/network/testdata/vlans.txt
@@ -1,0 +1,202 @@
+[!exec:bash] stop
+[!exec:grep] stop
+[!exec:cut] stop
+[!exec:curl] stop
+[!exec:sleep] stop
+
+# This test deploys 5 applications in total.
+# Four of those applications run the same curl and nginx services as used in "test_networking.txt".
+# All applications are connected to a NATed local network (10.1.0.0/24), used for access to HTTP endpoints from outside
+# through port-maps, but also to an air-gaped switch network split into multiple VLANs (10.2.<vid>.0/24):
+#  - app1 and app2 are connected to VLAN 100
+#  - app3 is connected to VLAN 200
+#  - app4 has initially no VLANs configured, later it is moved to VLAN 100
+#
+# The fifth application provides DHCP services separately for VLANs 100, 200 and for applications outside VLANs:
+#  - VLAN 100 uses IP range 10.2.100.0/24
+#  - VLAN 200 uses IP range 10.2.200.0/24
+#  - application interfaces without VLAN assignment get IP addresses from the range 10.2.0.0/24
+
+# The application providing DHCP services for multiple VLANs requires netadmin capabilities, otherwise
+# it is not permitted to create VLAN sub-interfaces. For security reasons, EVE does not allow to grant these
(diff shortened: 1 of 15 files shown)


mydatascience:
That's good enough. 
However it will be great if we get some test description in some table and put it in the docs section 
Our tests become complex. 
Also please add DHCP server deployment example to the apps list on wiki https://wiki.lfedge.org/display/EVE/Applications

tests/network/testdata/vlans.txt
```diff
@@ -0,0 +1,202 @@
...
+
+# The application providing DHCP services for multiple VLANs requires netadmin capabilities, otherwise
+# it is not permitted to create VLAN sub-interfaces. For security reasons, EVE does not allow to grant these
+# capabilities to native containers, therefore it is required to deploy apps as VMs-in-Containers for this test to pass.
+exec -t 2m bash check_vm_support.sh
+source .env
+[!env:with_hw_virt] skip 'Missing HW-assisted virtualization capability'
+
+# TODO: push into the itmoeve repository
+{{$image := "docker://milan4zededa/docker-test:1.3"}}
```
Insei, line 28: @milan-zededa We do not have a build for arm64 of this image, can we build it? otherwise rpi4 can no longer pass tests.

milan-zededa: Should be fixed now. Let me know if there are any further issues with this test on rpi4.

Insei: @milan-zededa Thanks for fixing it, but there is a problem with xen, and as far as I understand, there is a problem on x86 as well, since the logs are the same.
x86 XEN: https://github.com/lf-edge/eden/runs/2971302341?check_suite_focus=true
arm64 (rpi4) XEN: https://github.com/Insei/eden/runs/3003994600?check_suite_focus=true
Is there a way to fix this? Or are we only focused on kvm so far?

milan-zededa: Fix for Xen is on the way: https://github.com/lf-edge/eve/pull/2162

Insei: Thank you very much for the answer, did not see it.

PR #653, opened 2021-07-07
[WIP] Initial version cli interface for Arm Single Board Devices Deployment System demo api

Insei:
# About
I am developing a system that can automatically deploy operating systems and uninstall them on devices added to this system. We are talking about ARM single board devices similar to the Raspberry pi 4. Of the noteworthy, the system is also launched on rpi4.

This PR provides a cli interface to the API of this system.

System sources and readme about system: 
https://github.com/Insei/ASBDDS

# How to try
The system itself is only accessible after a VPN connection. This is a necessary level of security as long as the system itself does not have any means of authorization. Contact me for get access.

# Results
https://github.com/Insei/eden/actions/runs/1009249299 (EVE 6.9.0 RELEASE TEST)
https://github.com/Insei/eden/actions/runs/1006186194 (EVE 6.8.0 RELEASE TEST)

https://github.com/Insei/eden/blob/rpi4-auto-testing-demo/.github/workflows/rpi4.yml

# TODO
* Add more devices in pool (now i have only one rpi4 8gb in devices pool)
* Add workflow
* Write readme

### DIFF
--- a/cmd/edenAsbdds.go
+++ b/cmd/edenAsbdds.go
@@ -56,6 +56,41 @@ var asbddsDeviceDeleteCmd = &cobra.Command{
 		fmt.Println(string(jsonStr[:]))
 	},
 }
+
+var asbddsDeviceGetCmd = &cobra.Command{
+	Use:   "get",
+	Short: `get device`,
+	Run: func(cmd *cobra.Command, args []string) {
+		asbddsClient, err := linuxkit.NewASBDDSClient()
+		if err != nil {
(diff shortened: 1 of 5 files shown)


mydatascience:
Should we pull in the repo of https://github.com/Insei/ASBDDS inside? @Insei  Or at least make some consistent deployment enabled by eden. Otherwise we may end up with inconsistency and have problems with maintainability. We may rely on well-maintained project, but I'd vote to pull in other pieces.

PR #655, opened 2021-07-13
stringify the fields inside the attr field which were int and  boolean

chethan-zededa:
stringify the fields inside the attr field which were int and  boolean, since unmarshalling of these fields were resulting in errors

`json: cannot unmarshal number into Go value of type string`
few 'usagePolicy' was set to {} which was set to null

Signed-off-by: chethan-zededa <chethan@zededa.com>

### DIFF
--- a/models/template_l1_APU-2.json
+++ b/models/template_l1_APU-2.json
@@ -6,9 +6,9 @@
     "memory": "4G",
     "storage": "56G",
     "Cpus": "4",
-    "hsm": 0,
-    "leds": 3,
-    "watchdog": true
+    "hsm": "0",
+    "leds": "3",
+    "watchdog": "true"
   },
   "logo": {
     "logo_front": "/workspace/spec/logo_front_APU-2.png",
--- a/models/template_l1_Dell-Edge-Gateway-3001.json
+++ b/models/template_l1_Dell-Edge-Gateway-3001.json
@@ -4,9 +4,9 @@
     "memory": "4G",
     "storage": "32G",
     "Cpus": "4",
-    "hsm": 2,
-    "leds": 1,
-    "watchdog": true
+    "hsm": "2",
+    "leds": "1",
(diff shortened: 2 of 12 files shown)


eriknordmark (commented):
Two things:
1. Need to understand how to handle the transition with existing model files; If the user has saved a model file using zcli model export-manifest, can it be imported? How do we prevent the creating of a new enterprise to fail (when a set of default models are added)?
2. We have a script in EVE (pkg/debug/spec.sh) which generates in initial model file. That scipt needs to be changed.

Do we have a document or reference indicating why we need to change this stringification?

chethan-zededa:
> Two things:
> 
> 1. Need to understand how to handle the transition with existing model files; If the user has saved a model file using zcli model export-manifest, can it be imported? How do we prevent the creating of a new enterprise to fail (when a set of default models are added)?
> 2. We have a script in EVE (pkg/debug/spec.sh) which generates in initial model file. That scipt needs to be changed.
> 
> Do we have a document or reference indicating why we need to change this stringification?

it's been a while(like 2-3 months or more) zedcloud started jsonpb marshalling and unmarshalling, jsonpb operations are a bit strict wrt to these kind of type mismatch.
1. for an existing exported model file, it will continue to fail like today, even today if user tries create or update the model with type mismatch zedcloud will not process the request.
2. I was not aware of that script, let me take a look of it.

there is no document as such, as I said, model creates of such models were failing during enterprise onboard.
Screenshot 2021-07-13 at 9 10 01 PM

rvs:
> We have a script in EVE (pkg/debug/spec.sh) which generates in initial model file. That scipt needs to be changed.

That change just got merged @eriknordmark -- agree with the rest of your comments. Would love to hear from @chethan-zededa

eriknordmark (approved):
LGTM.
Thanks for the PR in EVE

PR #656, opened 2021-07-14
Test ACEs with the Drop action

milan-zededa:
PR adds support to define ACEs with the drop action and extends the `acl` test. Currently, the "Drop" action is not properly implemented by EVE (it behaves as "Allow").
Bug fix is provided by this PR: https://github.com/lf-edge/eve/pull/2184
Until the fix is merged and the eden is updated to test EVE with a tag recent enough to include the fix, this PR will keep failing and should not be merged, hence marking as Draft.

Also fixed is an occasional failure of the `port_forward` test. After a network outage (which is being exercised by that test), EVE is recovering some network configuration which was removed when the link went down. This for example includes iptables rules for portmaps. If eden tries to ssh too soon, it may actually get stucked, even if run with `ConnectTimeout=10`. This is probably because the timeout only applies to connection establishment, but not to subsequent communication, which apparently may get stuck if DNAT rules are being reinstalled at the same time. Using`timeout` command to limit the execution time of the `ssh` process as a whole is a completely reliable alternative.

### DIFF
--- a/cmd/edenPod.go
+++ b/cmd/edenPod.go
@@ -63,14 +63,23 @@ var podCmd = &cobra.Command{
 	Use: "pod",
 }
 
-func processAcls(acls []string) map[string][]string {
-	m := map[string][]string{}
+func processAcls(acls []string) expect.ACLs {
+	m := expect.ACLs{}
 	for _, el := range acls {
-		parsed := strings.SplitN(el, ":", 2)
+		parsed := strings.SplitN(el, ":", 3)
+		ni := parsed[0]
+		var ep string
 		if len(parsed) > 1 {
-			m[parsed[0]] = append(m[parsed[0]], parsed[1])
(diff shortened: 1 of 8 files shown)


rvs:
Is this ready for review yet?

milan-zededa:
> Is this ready for review yet?

For review it is ready. Not for merge though. Need [this](https://github.com/lf-edge/eve/pull/2184) to get into EVE first, hence the PR is marked as draft for now.

rvs (approved):
LGTM. @giggsoff ?

giggsoff (approved):
Well written, thanks!

rvs:
So the code is in EVE -- seems like we can merge this?

milan-zededa:
> So the code is in EVE -- seems like we can merge this?

yes, ready for merge

PR #665, opened 2021-08-10
Add model information for OnLogic CL210G-11-RS

floucks:
@rvs as promised, here is the pull request for adding support for the CL210G-11-RS. 
I've added the template file (including image names & product url), uploaded image files, and updated the README. 

Let me know if anything else is needed. Happy to help!


floucks:
I followed the instructions from the DCO workflow regarding proper commit signing and have pushed back to master on my fork. LMK if I need to close / re-submit the PR or something.

rvs:
Perfect! Thank you so much @floucks -- this looks great -- merging it now.

eriknordmark (commented):
@rvs looks like this script output isn't valid json (due to a bug in spec.sh I suspect).

rvs:
Ah! True. Which remind me -- that we need at least a basic sanity check as part of Eden makefile -- let me fix that. Also, when you say a bug in spec.sh is that a bug you fixed recently?

eriknordmark:
> Ah! True. Which remind me -- that we need at least a basic sanity check as part of Eden makefile -- let me fix that. Also, when you say a bug in spec.sh is that a bug you fixed recently?

Bug was introduced in https://github.com/lf-edge/eve/commit/5341cc3baf78247e69691ed3f11c9a54a73421d4#diff-5082c5ce36254a19dd88fff26f3d05bc815ca75f4efb1a8b0d4868803ac6da51
Fix is sitting in https://github.com/lf-edge/eve/pull/2208 (which is waiting for me to complete the update of the markdown)

I don't know if we can have a test where Eden at least "imports" the model. Zedcloud for instance validates that there are no unknown fields in the json in addition to basic json parsing.

PR #749, opened 2022-03-03
Makefile: Make the tests target to depend on build-tests

lmr:
In order to run the tests target, we need to have the
test binaries built. Let's update that target to do
that beforehand.

This prevents situations where newcomers scratch their
heads on why the tests target fails and it doesn't seem
to imply any side-effects.

Signed-off-by: Lucas Meneghel Rodrigues <lucas@zededa.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ $(BINDIR):
 $(DIRECTORY_EXPORT):
 	mkdir -p $@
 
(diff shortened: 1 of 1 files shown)


giggsoff (approved):
LGTM

rvs (commented):
LGTM

PR #757, opened 2022-04-21
Test for location info reporting

milan-zededa:
This test partially covers functionality described here:
https://wiki.lfedge.org/display/EVE/GPS+coordinates+exposed+by+EVE


eriknordmark (commented):
Looks like some test issue:
cat: can't open '/mnt/location.json': No such file or directory

milan-zededa:
> Looks like some test issue:
> cat: can't open '/mnt/location.json': No such file or directory

Yes, that will be failing until we merge GPS PR into eve. That's why this is marked only as a draft for the time being.

giggsoff (approved):
In general it is not a good idea to feed values via ssh into managers, but seems there are no other approaches right now

PR #758, opened 2022-05-05
Add local profile server devinfo test

eriknordmark:
Tests for shutdown and shutdown_poweroff in the Local Profile Server DevInfo API while checking that an app
instance running a local profile server is shut down last.

Separate commit for a test for shutdown from controller.

Depends on the implementation in https://github.com/lf-edge/eve/pull/2610

### DIFF
--- a/tests/eclient/testdata/dev_local_info.txt
+++ b/tests/eclient/testdata/dev_local_info.txt
@@ -1,0 +1,258 @@
+# Test dev local info
+
+{{define "mngr_port"}}8027{{end}}
+{{define "app_port"}}8028{{end}}
+{{define "token"}}server_token_123{{end}}
+{{define "app_info_status_file"}}/mnt/app-info-status.json{{end}}
+{{define "dev_info_status_file"}}/mnt/dev-info-status.json{{end}}
+{{define "dev_cmd_file"}}/mnt/dev-command.json{{end}}
+{{define "network"}}n1{{end}}
+{{define "ssh"}}ssh -q -o ConnectTimeout=10 -o StrictHostKeyChecking=no -o PasswordAuthentication=no -i {{EdenConfig "eden.tests"}}/eclient/image/cert/id_rsa root@{{end}}
+
+[!exec:bash] stop
+[!exec:sleep] stop
+[!exec:ssh] stop
+[!exec:chmod] stop
+[!exec:jq] stop
+
+exec chmod 600 {{EdenConfig "eden.tests"}}/eclient/image/cert/id_rsa
+
+# Starting of reboot detector with a 3 reboot limit since we reboot twice
+! test eden.reboot.test -test.v -timewait 100m -reboot=0 -count=3 &
+
+message 'Resetting of EVE'
+eden eve reset
+exec sleep 30
+
+# Create n1 network
+eden -t 1m network create 10.11.12.0/24 -n {{template "network"}}
+test eden.network.test -test.v -timewait 10m ACTIVATED {{template "network"}}
+
+# Deploy local-manager
+eden pod deploy -n local-manager --memory=512MB docker://lfedge/eden-eclient:99d7f62 -p {{template "mngr_port"}}:22 --networks={{template "network"}}
+test eden.app.test -test.v -timewait 10m RUNNING local-manager
+
+# Wait for ssh access
+exec -t 5m bash wait-ssh.sh {{template "mngr_port"}}
+
+# Start local manager application
+exec -t 1m bash local-manager-start.sh
+
+# Obtain local-manager IP address
+exec -t 2m bash get-app-ip.sh local-manager
+source .env
+
+# Configure local server
+eden controller edge-node update --device profile_server_token={{template "token"}}
+eden controller edge-node update --device local_profile_server=$app_ip:8888
+
+# STEP 1: Wait for devinfo status
+exec sleep 30
+exec -t 10m bash get-devinfo-status.sh
+! stderr .
+
+# STEP 2: Deploy the second app
+eden pod deploy -n app1 --memory=512MB docker://lfedge/eden-eclient:9455582 -p {{template "app_port"}}:22 --networks={{template "network"}}
+test eden.app.test -test.v -timewait 10m RUNNING app1
+
+# Wait for ssh access
+exec -t 5m bash wait-ssh.sh {{template "app_port"}}
+
+# STEP 3: Wait for devinfo status reporting ONLINE
+exec -t 10m bash wait-for-dev-state.sh ONLINE
+! stderr .
+
+# STEP 4: Request for shutdown. Check that app1 stops before local_manager
+exec -t 1m bash put-devinfo-cmd.sh COMMAND_SHUTDOWN &
+exec -t 5m bash wait-for-dev-state.sh SHUTTING_DOWN
+! stderr .
+# Could it already have HALTED? Match HALTING or HALTED
+exec -t 5m bash wait-for-app-state.sh app1 HALT
+! stderr .
+exec -t 1m bash get-appinfo-status.sh local-manager
+stdout 'RUNNING'
+! stderr .
+exec -t 5m bash wait-for-app-state.sh app1 HALTED
+! stderr .
+
+# Did local-manager halt? Can't check by asking local-manager
+exec -t 1m bash eden-pod-ps.sh local-manager
+! stdout 'RUNNING'
+! stderr .
+
+# STEP 5: test reboot via controller bringing them back up
+# send reboot command without wait
+eden controller edge-node reboot
+
+# STEP 5.1: Wait for ssh access
+exec -t 40m bash wait-ssh.sh {{template "app_port"}}
+
+# STEP 5.2: Start local manager application
+exec -t 1m bash local-manager-start.sh
+
+# STEP 5.3: Wait for devinfo status reporting ONLINE
+exec sleep 30
+exec -t 10m bash wait-for-dev-state.sh ONLINE
+! stderr .
+
+# STEP 5.4: Check apps are back
+exec -t 20m bash wait-for-app-state.sh local-manager RUNNING
+! stderr .
+exec -t 20m bash wait-for-app-state.sh app1 RUNNING
+! stderr .
+
+# We can run the rest of the test only with qemu right now to properly
+# handle start of EVE after poweroff.
+{{$devmodel := EdenConfig "eve.devmodel"}}
+{{if not (eq $devmodel "ZedVirtual-4G")}}
+eden eve reset
+skip 'The rest of the test is supported only on QEMU'
+{{end}}
+
+# STEP 6: Request for poweroff. Check that app1 stops before local_manager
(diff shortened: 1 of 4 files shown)


tests/eclient/testdata/dev_local_info.txt
```diff
@@ -0,0 +1,241 @@
...
+message 'Resetting of EVE'
+eden eve reset
+exec sleep 30
+
+# Create n1 network
+eden -t 1m network create 10.11.12.0/24 -n {{template "network"}}
+test eden.network.test -test.v -timewait 10m ACTIVATED {{template "network"}}
+
+# Deploy local-manager
+eden pod deploy -n local-manager --memory=512MB docker://lfedge/eden-eclient:afaa332 -p {{template "mngr_port"}}:22 --networks={{template "network"}}
```
giggsoff, line 32: Version of eclient should be updated after image push, and we push images on PR merge right now. So please split PR onto two: one for update of eclient image and the second one for update tests.

tests/eclient/testdata/dev_local_info.txt
```diff
@@ -0,0 +1,241 @@
...
+
+# Obtain local-manager IP address
+exec -t 2m bash get-app-ip.sh local-manager
+source .env
+
+# Configure local server
+eden controller edge-node update --device profile_server_token={{template "token"}}
+eden controller edge-node update --device local_profile_server=$app_ip:8888
+
+# STEP 1: Wait for devinfo status
```
giggsoff, line 49: I think it will be nice to add sleep for i.e. 30 seconds before this step.

tests/eclient/testdata/dev_local_info.txt
```diff
@@ -0,0 +1,241 @@
...
+! stderr .
+
+# STEP 7: Check qemu process is gone aka powered off
+exec sleep 120
+eden eve status
+! stdout process not running'
+
+# STEP 8: Restart EVE; check apps come up
+message 'Resetting of EVE'
+eden eve reset
```
giggsoff, line 118: It should be `eden eve start`. Reset command will remove all apps from EVE configuration but will not affect qemu process state. In general seems we can run this only for qemu (see another comment).

tests/eclient/testdata/dev_local_info.txt
```diff
@@ -0,0 +1,241 @@
...
+# STEP 9: Remove the second app
+eden pod delete app1
+test eden.app.test -test.v -timewait 15m - app1
+
+# STEP 10: Undeploy local-manager
+eden pod delete local-manager
+test eden.app.test -test.v -timewait 15m - local-manager
+eden network delete {{template "network"}}
+test eden.network.test -test.v -timewait 10m - {{template "network"}}
+
```
giggsoff, line 142: We should put `eden eve reset` and small sleep here to remove configuration of LPS from EVE config.

tests/eclient/testdata/dev_local_info.txt
```diff
@@ -0,0 +1,241 @@
...
+-- put-devinfo-cmd.sh --
+DN="$1"
+TIMESTAMP="${2:-0}"
+CMD="${3:-COMMAND_UNSPECIFIED}"
+
+EDEN={{EdenConfig "eden.root"}}/{{EdenConfig "eden.bin-dist"}}/{{EdenConfig "eden.eden-bin"}}
+HOST=$($EDEN eve ip)
+CONFIG="{\"displayname\": \"$DN\", \"timestamp\": $TIMESTAMP, \"command\": \"$CMD\"}"
+echo "$CONFIG" | {{template "ssh"}}$HOST -p {{template "mngr_port"}} 'cat > {{template "dev_cmd_file"}}'
```
giggsoff, line 219: As I understand it should be json representation of [LocalDevCmd](https://github.com/lf-edge/eve/blob/master/api/go/profile/local_profile.pb.go#L815-L826), so command is enough. Also I can see only one argument for calling this function, so please remove DN and TIMESTAMP.

tests/eclient/testdata/dev_local_info.txt
```diff
@@ -0,0 +1,241 @@
...
+
+# STEP 5: test reboot via controller bringing them back up
+# send reboot command and wait in background
+test eden.reboot.test -test.v -timewait=20m -reboot=1 -count=1 &
+exec -t 5m bash wait-for-app-state.sh local-manager RUNNING
+! stderr .
+exec -t 5m bash wait-for-app-state.sh app1 RUNNING
+! stderr .
+
+# STEP 6: Request for poweroff. Check that app1 stops before local_manager
```
giggsoff, line 93: We can run the rest of the test only with qemu right now to properly handle start of EVE after poweroff.
Please add the following before the step 6:
```
{{$devmodel := EdenConfig "eve.devmodel"}}
{{if not (eq $devmodel "ZedVirtual-4G")}}
eden eve reset
skip 'The rest of the test is supported only on QEMU'
{{end}}
```

tests/eclient/testdata/dev_local_info.txt
```diff
@@ -0,0 +1,241 @@
...
+[!exec:bash] stop
+[!exec:sleep] stop
+[!exec:ssh] stop
+[!exec:chmod] stop
+[!exec:jq] stop
+
+exec chmod 600 {{EdenConfig "eden.tests"}}/eclient/image/cert/id_rsa
+
+# Starting of reboot detector with a 2 reboot limit
+! test eden.reboot.test -test.v -timewait 100m -reboot=0 -count=2 &
```
giggsoff, line 21: It should catch 3 reboots to fail (2 reboots are expected, the third one is an error).

giggsoff (changes requested):
Please put eclient image update into the separate PR and see comments

tests/eclient/testdata/dev_local_info.txt
eriknordmark: Ah - that docker tag is what selects the version. Got it.

eriknordmark: Created a separate https://github.com/lf-edge/eden/pull/762 with eclient.

tests/eclient/testdata/dev_local_info.txt
eriknordmark: OK

tests/eclient/testdata/dev_local_info.txt
eriknordmark: OK

tests/eclient/testdata/dev_local_info.txt
eriknordmark: Before the pod delete? Or before the delete of the local-manager?

I don't think we have that step in the other uses of the local profile server; I copied this from app_local_info.txt

tests/eclient/testdata/dev_local_info.txt
eriknordmark: Thanks

tests/eclient/testdata/dev_local_info.txt
eriknordmark: OK, didn't realize the limit had that semantics; thought it was the expected number not the number at which it is a failure.

tests/eclient/testdata/dev_local_info.txt
giggsoff: Timestamp set automatically based on file modification time here: https://github.com/lf-edge/eden/blob/99d7f6244a1b98b3fb1b3372f0346bb44babebe1/tests/eclient/image/pkg/main.go#L219

tests/eclient/testdata/dev_local_info.txt
giggsoff: Oh, seems I missed line here, after waiting for network delete, in the end.
Also please update eve version to one with PR merged here https://github.com/lf-edge/eden/blob/99d7f6244a1b98b3fb1b3372f0346bb44babebe1/pkg/defaults/defaults.go#L53

tests/eclient/testdata/dev_local_info.txt
eriknordmark: Right. Fixing that.

tests/eclient/testdata/dev_local_info.txt
eriknordmark: When do we update DefaultEVETag? Any time there is a new feature in EVE-OS needed to pass the tests?

giggsoff: Yes, we update version inside Eden repo when we have new feature in EVE test need to run. Inside EVE repo we use current (the last) version of EVE for tests.

tests/eclient/testdata/dev_local_info.txt
```diff
@@ -0,0 +1,262 @@
...
+eden controller edge-node reboot
+
+# STEP 5.1: Wait for ssh access
+exec -t 40m bash wait-ssh.sh {{template "app_port"}}
+
+# STEP 5.2: Start local manager application
+exec -t 1m bash local-manager-start.sh
+
+# STEP 5.3: Wait for new devinfo status
+exec sleep 30
```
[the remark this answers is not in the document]
eriknordmark: But we can't wait for the devinfo as reported by the local-manager since it isn't running. Do we have some function to look at the relatively new state reported by the device itself in ZDeviceState?

One thing I need to check is how EVE-OS decides to back-off the LPS API calls. It might need some patience after a device reboot since the LPS might not be running yet.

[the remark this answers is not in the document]
eriknordmark: @giggsoff I don't see any match on ZDeviceState in eden source code.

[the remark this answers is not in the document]
giggsoff: I mean you can call script `wait-for-dev-state.sh` with ONLINE match instead of `get-devinfo-status.sh`. With this change we will use loop and will check the file inside LPS several times.

tests/eclient/testdata/shutdown_test.txt
```diff
@@ -0,0 +1,73 @@
...
+# stdout 'Ubuntu'
+
+# send shutdown command
+eden controller edge-node shutdown
+
+# wait for HALTED state which indicates that we are shutting down
+test eden.app.test -test.v -timewait 5m HALTED {{$app_name}}
+# XXX can we check for the SHUTTING_DOWN state for the edge-node?
+
+# now reboot node to bring app back up
```
giggsoff, line 34: Please sync timewaits inside this block. Timewait for app and network should be more than timewait for reboot.

tests/eclient/testdata/dev_local_info.txt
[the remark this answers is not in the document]
eriknordmark: Sure I can change that (I thought this check was before the LPS was started but it is after it is running.)

tests/eclient/testdata/shutdown_test.txt
eriknordmark: Updated

giggsoff (approved):
LGTM

will work on SHUTTING_DOWN state check in next PR

PR #762, opened 2022-05-12
Add /api/v1/devinfo to eclient

eriknordmark:
This is a new API endpoint for the Local Profile Server enabling
shutdown of all apps and optionally power off the device. It also
reports the device state like maintenance mode.


giggsoff (approved):
LGTM

PR #820, opened 2022-11-29
Avoid confusion between EveConfigDir and ConfigDist options

milan-zededa:
Because of a mixup between `EveConfigDir` and previously called `ConfigDir`, `eve-config-dir` used with `eden setup` option would not work.


giggsoff (approved):
Thank you for finding and fixing it!

PR #851, opened 2023-03-17
Fix "controller edge-node set-config" command

milan-zededa:
In [pkg/controller/adam/adam.go](https://github.com/lf-edge/eden/blob/master/pkg/controller/adam/adam.go), method [ConfigSet](https://github.com/lf-edge/eden/blob/master/pkg/controller/adam/adam.go#L190-L193), we have recently changed content type to `application/x-proto-binary`, yet [EdgeNodeSetConfig](https://github.com/lf-edge/eden/blob/master/pkg/openevec/edgeNode.go#L303-L307) remained unchanged and still submits the configuration in JSON.


giggsoff (approved):
Thank you for finding and fixing it!

PR #858, opened 2023-04-04
sdn,client: use the authentication identity files only for the ssh

rouming:
By default SSH tries to access the server with all the keys from the ~/.ssh folder, and the identity will be the last one in the list. This behavior leads to the following error:

    ERRO[0000] Failed to get EVE IP address: get-eve-ip.sh failed: Received disconnect from 127.0.0.1 port 6622:2: Too many authentication failures

if there are number of keys in the ~/.ssh folder is greater than 'MaxAuthTries' SSH server configuration option, which is set to 6 by default.

сс: @milan-zededa 
cc: @uncleDecart

### DIFF
--- a/pkg/edensdn/client.go
+++ b/pkg/edensdn/client.go
@@ -171,7 +171,7 @@ func (client *SdnClient) sshArgs(extra ...string) (sshArgs []string) {
 	if client.SSHKeyPath == "" {
 		log.Fatal("SDN client with undefined SSHKeyPath")
 	}
-	allArgs := fmt.Sprintf("-o ConnectTimeout=5 -o StrictHostKeyChecking=no "+
+	allArgs := fmt.Sprintf("-o IdentitiesOnly=yes -o ConnectTimeout=5 -o StrictHostKeyChecking=no "+
 		"-i %s -p %d root@localhost", client.SSHKeyPath, client.SSHPort)
 	return append(strings.Fields(allArgs), extra...)
 }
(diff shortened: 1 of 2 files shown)


pkg/edensdn/client.go
```diff
@@ -171,7 +171,7 @@ func (client *SdnClient) sshArgs(extra ...string) (sshArgs []string) {
 	if client.SSHKeyPath == "" {
 		log.Fatal("SDN client with undefined SSHKeyPath")
 	}
-	allArgs := fmt.Sprintf("-o ConnectTimeout=5 -o StrictHostKeyChecking=no "+
+	allArgs := fmt.Sprintf("-o IdentitiesOnly=yes -o ConnectTimeout=5 -o StrictHostKeyChecking=no "+
 		"-i %s -p %d root@localhost", client.SSHKeyPath, client.SSHPort)
```
milan-zededa, line 175: Is the `-i %s` parameter not preventing ssh to look for other identities?

rouming: Nope, unfortunately. See the debug output of the ssh:

```
debug1: Will attempt key: ro@dl RSA SHA256 agent
debug1: Will attempt key: ro@dl RSA SHA256 agent
debug1: Will attempt key: ro@p RSA SHA256 agent
debug1: Will attempt key: ro@fi RSA SHA256 agent
debug1: Will attempt key: kn@mi RSA SHA256 agent
debug1: Will attempt key: sss RSA SHA256 agent
debug1: Will attempt key: /eden/sdn/vm/cert/ssh/id_rsa RSA explicit
```
the eden key is at the end. 

with the `IdentitiesOnly=yes` ssh will try only identity, skipping other keys

milan-zededa: OK, good to know

milan-zededa (approved):

uncleDecart (approved):
LGTM

PR #861, opened 2023-05-02
Add StopSDN call lost during refactoring

milan-zededa:
As most functions have moved to `openevec` package, `StopEve()` accidentally lost call to `StopSDN()`, which stops SDN VM if it is running.
@uncleDecart, I noticed that `StopEVE()` is duplicated and still exists also outside of `openevec`: https://github.com/lf-edge/eden/blob/master/pkg/eden/eden.go#L896-L926

### DIFF
--- a/pkg/openevec/eve.go
+++ b/pkg/openevec/eve.go
@@ -210,6 +210,7 @@ func StopEve(vmName string, cfg *EdenSetupArgs) error {
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


milan-zededa:
> What do you think if we call eden.StopEve from pkg/openevec for now?

Sorry, misunderstood the question.
Yes that could work as well.

uncleDecart (approved):
LGTM. Let's create a ticket for clean-up, but what I meant is calling eden.StopEVE from pkg/openevec/eve.go instead of StopEden

PR #865, opened 2023-05-10
Small fixes for unstable eden tests

milan-zededa:
Mostly just increasing timeouts where we often see EVE not being fast enough when run by slow Github runners.


uncleDecart (approved):
LGTM, 

We might want to make this waiting parameters dependent on platform they are running, but since they don't slow us down I think it's alright atm

PR #879, opened 2023-07-24
use actual value of the repo for the checkout 

yash-zededa:
`${{ github.event.repository.full_name }}` is not replaced by when the WF is called from another workflow. Use the actual value for the repo checkout for actions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Get code
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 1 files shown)


yash-zededa:
@uncleDecart Let's keep this PR open till the time we have tested the workflow call from eve. Testing under my fork repo's

yash-zededa:
@uncleDecart Let's merge this.

uncleDecart (approved):
LGTM

PR #880, opened 2023-07-25
Traffic control for Eden-SDN

milan-zededa:
With this patch, Eden-SDN now allows to apply traffic control individually for every network port. This can be used to model poor network connectivity and observe how EVE is able to deal with such challenging conditions.


uncleDecart (approved):
LGTM, but there's some weird failure with `setup_eden`. Worker runs out of space, most likely it's not because of this PR, but I suggest we finish all our shenanigans with workflows before merging this @milan-zededa 

And thank you for this test Milan! :art:

milan-zededa:
OK, let's finish the shenanigans first :)

PR #881, opened 2023-07-27
setting eve.tag and eve.registry for eden tests for PR's build

yash-zededa:
A different repository is used for the eve PRs after they have been built. Although this value is present in the tag that Eve passes, we still need to extract the necessary data. Using the fundamental shell commands, this PR will retrieve the necessary data.

### DIFF
--- a/.github/actions/setup-environment/action.yml
+++ b/.github/actions/setup-environment/action.yml
@@ -8,7 +8,7 @@ inputs:
   tpm_enabled:
     required: true
     type: bool
-  eve_tag:
+  eve_image:
     type: string
 
 runs:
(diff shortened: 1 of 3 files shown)


.github/actions/setup-environment/action.yml
```diff
@@ -50,9 +50,13 @@ runs:
       shell: bash
       working-directory: "./eden"
     - name: Setup eve version
-      if: "${{ inputs.eve_tag != '' }}"
+      if: ${{ inputs.eve_tag != '' }} and contains(${{ inputs.eve_tag }}, ":" )
```
uncleDecart, line 53: That means we will just ignore any input which does not contain `:` which might be misleading because input variable name is `eve_tag`, I'd put this check inside `run:` section that way this action can use eve_tag which could be just tag or contain registry. Because this way if you want to run eden tests for specific tag inside eden repository you'll have to specify lf-edge/eden:$tag$ as eve tag which IMO redundant. 
Either this or changing name to `eve_image`

uncleDecart (commented):
LGTM except confusion with `eve_tag` forcing to have `:`

eriknordmark:
I see a no space left on device in the failure

eriknordmark:
> That's interesting @eriknordmark , because we are still not using this workflows, we run them manually till they are successful (and only then we want to use them in PRs). So this patch does not affect other workflows running (including eden_setup)

I was referring to the error for this PR in https://github.com/lf-edge/eden/actions/runs/5681201583/job/15396839195?pr=881

And they are being used in EVE (and all failing). See https://github.com/lf-edge/eve/actions/workflows/eden_test.yml
I assumed that was known

.github/actions/setup-environment/action.yml
yash-zededa: Will change the inputs to `eve_image` to be more consistent.

yash-zededa:
> Reference

Yes, the `Test from lf-edge/eden` WF is being used by eve to test the reusable workflow functionality. Once this PR is merged. I will raise a PR that will fix eve.

uncleDecart (approved):
LGTM

PR #883, opened 2023-07-31
Add missing `working-directory` tag in `run-eden-test` action

uncleDecart:
Since we are using `test.yml` as reusable workflow in EVE we clone eden repository in EVE folder and it causes problems with working directories and building and executing binaries.

So we are using `working-directory:` and `path:` tags of GitHub actions with hard-coded `./eden` value to prevent any problems that might occur with resuing this workflow in EVE.

### DIFF
--- a/.github/actions/run-eden-test/action.yml
+++ b/.github/actions/run-eden-test/action.yml
@@ -30,12 +30,14 @@ runs:
     - name: Collect info
       if: failure()
       uses: ./eden/.github/actions/collect-info
+      working-directory: "./eden"
     - name: Collect logs
       if: always()
       uses: ./eden/.github/actions/publish-logs
       with:
         file_system: ${{ inputs.file_system }}
         tpm_enabled: ${{ inputs.tpm_enabled }}
+      working-directory: "./eden"
     - name: Clean up after test
       if: always()
       run: |


uncleDecart:
During running https://github.com/lf-edge/eden/actions/runs/5712658495/job/15476576181 there was an error with collect-info action, this PR should fix it

cc: @milan-zededa @yash-zededa @giggsoff

milan-zededa (approved):

uncleDecart:
@eriknordmark can you please review and approve this?

giggsoff (approved):
Hope we will refactor the place of setting this option later (for example we set it inside of action [here](https://github.com/lf-edge/eden/blob/dee1c3654e1503a642353d42b2e2b871659cedd0/.github/actions/setup-environment/action.yml#L33) ).

uncleDecart:
Problem is that as I understand is that globally setting it won't set it locally, but @yash-zededa can correct me if I'm wrong

yash-zededa:
> Problem is that as I understand is that globally setting it won't set it locally, but @yash-zededa can correct me if I'm wrong

That's correct. I have tried to set this globally inside the actions and for me it was not working. Seems like the `defaults` option is only supported for workflow `run` commands

PR #892, opened 2023-09-28
Add CODEOWNERS file`

uncleDecart:
I sometimes miss PRs for eden, so I'm adding myself as code owner, so that I'll get notifications when somebody opens PR

cc: @eriknordmark

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,5 @@
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence, these will
+# be requested for review when someone opens a pull request.
+
+*       @uncleDecart


milan-zededa:
Github allows to subscribe for all activity on the repository ;)

uncleDecart:
That was lightning fast :D But that won't add me as reviewer on PR, would it?

milan-zededa:
> That was lightning fast :D But that won't add me as reviewer on PR, would it?

No, but you will not miss anything.

uncleDecart:
Fair @milan-zededa , but I think adding CODEOWNERS make sense to automatically assign PR on eden, right now it could be simple * pattern for now, but it's better to start early, than late. It is easier to add small things, than to try to fix big one

For instance, we can assign sdn on you, since you're the author :D And I can also add you under * pattern

PR #895, opened 2023-10-03
Collect EVE info via console if ssh access is not working

milan-zededa:
We use [collect-info](https://github.com/lf-edge/eve/blob/master/pkg/debug/scripts/collect-info.sh) script to take a snapshot of the EVE state when tests fail and publish them alongside logs. Previously, we relied on the ssh access to produce and download EVE info tarball. However, this is not available when device is failing to onboard or if there are some networking issues. In that case, we can make (a rather desperate) attempt to get debug info over the serial console.

### DIFF
--- a/.github/actions/collect-info/action.yml
+++ b/.github/actions/collect-info/action.yml
@@ -1,20 +1,14 @@
 name: 'Collect and store debug info'
-description: 'Collect debug info using EVE script executed via ssh and store downloaded tarball under the specified file name'
+description: 'Collect debug info using EVE script executed via ssh or console and store downloaded tarball under the specified file name'
 
 runs:
   using: 'composite'
   steps:
     - name: Collect info
       run: |
-        # Give EVE 5 minutes at most to enable ssh access (if tests failed early).
-        for i in $(seq 60); do ./eden eve ssh && break || sleep 5; done
-        ./eden sdn fwd eth0 22 --\
-          ssh -o StrictHostKeyChecking=no -p FWD_PORT -i ./dist/default-certs/id_rsa root@FWD_IP collect-info.sh &&\
-        ./eden sdn fwd eth0 22 --\
-          scp -o StrictHostKeyChecking=no -P FWD_PORT -i ./dist/default-certs/id_rsa root@FWD_IP:/persist/eve-info-* . &&\
-        # upload-artifact complains about colon in the file name
-        # make sure to update upload step if changing name
-        mv eve-info-* eve-info.tar.gz ||\
-        echo "failed to collect info"
+        # Do not pollute console logs which are collected by publish-logs action.
+        cp dist/default-eve.log dist/default-eve.log.backup || true
+        ./shell-scripts/collect-info-ssh.sh || ./shell-scripts/collect-info-console.sh 120
+        cp dist/default-eve.log.backup dist/default-eve.log || true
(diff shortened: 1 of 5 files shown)


uncleDecart (commented):
I suggest we move shell-scripts to actions/collect-info and use this action in github/workflows/eden as well. Also don't forget backslashes in running command :)

milan-zededa:
> I suggest we move shell-scripts to actions/collect-info and use this action in github/workflows/eden as well. Also don't forget backslashes in running command :)

I did put this to both `.github/actions/collect-info/action.yml` and to `.github/workflows/eden.yml`.
These shell scripts can be used also manually when running eden locally, not just from actions. That's why I put them to `shell-scripts` and not under `github/actions`.
Where exactly do I miss backslashes?

.github/actions/collect-info/action.yml
```diff
@@ -1,20 +1,14 @@
...
-        ./eden sdn fwd eth0 22 --\
-          ssh -o StrictHostKeyChecking=no -p FWD_PORT -i ./dist/default-certs/id_rsa root@FWD_IP collect-info.sh &&\
-        ./eden sdn fwd eth0 22 --\
-          scp -o StrictHostKeyChecking=no -P FWD_PORT -i ./dist/default-certs/id_rsa root@FWD_IP:/persist/eve-info-* . &&\
-        # upload-artifact complains about colon in the file name
-        # make sure to update upload step if changing name
-        mv eve-info-* eve-info.tar.gz ||\
-        echo "failed to collect info"
+        # Do not pollute console logs which are collected by publish-logs action.
+        cp dist/default-eve.log dist/default-eve.log.backup || true
```
uncleDecart, line 10: `cp dist/default-eve.log dist/default-eve.log.backup || true \\`
and more below

milan-zededa: Is that really necessary? We do not use backslashes in other places, e.g.: https://github.com/lf-edge/eden/blob/master/.github/workflows/eden.yml#L75-L80

uncleDecart:
>> I suggest we move shell-scripts to actions/collect-info and use this action in github/workflows/eden as well. Also don't forget backslashes in running command :)

>I did put this to both .github/actions/collect-info/action.yml and to .github/workflows/eden.yml.
These shell scripts can be used also manually when running eden locally, not just from actions. That's why I put them to shell-scripts and not under github/actions.
Where exactly do I miss backslashes?

My bad, did not parse file-tree correctly :D, is it useful to run those scripts locally?

milan-zededa:
> > > I suggest we move shell-scripts to actions/collect-info and use this action in github/workflows/eden as well. Also don't forget backslashes in running command :)
> 
> > I did put this to both .github/actions/collect-info/action.yml and to .github/workflows/eden.yml.
> > These shell scripts can be used also manually when running eden locally, not just from actions. That's why I put them to shell-scripts and not under github/actions.
> > Where exactly do I miss backslashes?
> 
> My bad, did not parse file-tree correctly :D, is it useful to run those scripts locally?

These scripts are also helpful to quickly collect info when debugging something locally.

.github/actions/collect-info/action.yml
uncleDecart: I guess we can skip it here, do you want to print something in case cp fails?

PR #897, opened 2023-10-04
GitHub Actions: use report_name instead of variables in publish-logs

uncleDecart:
Previous implementation did not use test suite name in publishing logs, therefore for 11 different workflows we got only 4 artifacts, new implimentation uses report_name parameter in publish-logs action which makes it more flexible to set up any report name user wants and makes it more clear how it forms in `run-eden-test` action. And, of course, it adds test suite name now

cc: @milan-zededa @yash-zededa 

Please approve: @deitch @eriknordmark

### DIFF
--- a/.github/actions/run-eden-test/action.yml
+++ b/.github/actions/run-eden-test/action.yml
@@ -34,8 +34,7 @@ runs:
       if: always()
       uses: ./eden/.github/actions/publish-logs
       with:
-        file_system: ${{ inputs.file_system }}
-        tpm_enabled: ${{ inputs.tpm_enabled }}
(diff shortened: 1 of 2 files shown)


.github/actions/run-eden-test/action.yml
```diff
@@ -34,6 +34,7 @@ runs:
       if: always()
       uses: ./eden/.github/actions/publish-logs
       with:
+        report_name: eden-report-${{ inputs.suite }}-tpm-${{ inputs.tpm_enabled }}-${{ inputs.file_system }}
         file_system: ${{ inputs.file_system }}
```
milan-zededa, line 38: file_system and tpm_enabled can be removed from "with"

uncleDecart: fixed

eriknordmark (approved):
LGTM

PR #905, opened 2023-10-12
Test DNS resolution from apps

milan-zededa:
This test checks that apps connected to the same local network instance are able to resolve each other's IP addresses from their display names using the DNS service of the network instance.
Also tested is resolution of static DNS entries configured for the network instance.

Note that there was a bug affecting DNS resolution for apps with multiple interfaces in the same network instance. This was fixed in 10.11, meaning that this is the minimal version for this test to pass.

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -53,7 +53,7 @@ const (
 	DefaultRegistryPort         = 5050
 
 	//tags, versions, repos
-	DefaultEVETag               = "10.10.0" // DefaultEVETag tag for EVE image
+	DefaultEVETag               = "10.11.0" // DefaultEVETag tag for EVE image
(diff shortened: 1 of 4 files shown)


eriknordmark (commented):
Unrelated to this PR, we need to make eden run the new decomposed eden test workflow as we now do in the eve repo.

milan-zededa:
> Unrelated to this PR, we need to make eden run the new decomposed eden test workflow as we now do in the eve repo.

@eriknordmark Added [comment to the respective issue](https://github.com/lf-edge/eden/issues/863#issuecomment-1761314249).

milan-zededa:
@uncleDecart @giggsoff This test is also ready for review (passes on the latest EVE release).

giggsoff (approved):

PR #908, opened 2023-10-19
Remove config check

uncleDecart:


uncleDecart:
This is prerequisite to use edenSetup and openevec in general, to use it without file

PR #914, opened 2023-10-27
Revert "pkg: defaults: Workaround to fix QEMU's MSR errors"

rene:
This reverts commit 72eb8206f5987c544fec30d5dc33bb5df4076eb5.

It seems this workaround is no longer required on newer systems and it's causing a new error:

KVM: entry failed, hardware error 0x80000021

Removing the workaround solves this issue and it also works with kernel 5.10.186.


eriknordmark (approved):
LGTM

PR #917, opened 2023-11-03
add security test suite to eden

shjala:
This adds security test suite to eden. As suggested by @uncleDecart, to liberate ourselves from bash and eden script, it is written in go ;)

For now it just check apparmor enablement, more to be added to the suite.

### DIFF
--- a/tests/sec/sec_test.go
+++ b/tests/sec/sec_test.go
@@ -1,0 +1,110 @@
+package sec_test
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/lf-edge/eden/pkg/device"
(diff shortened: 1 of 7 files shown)


uncleDecart (approved):
LGTM, thank you @shjala for this!

cc: @milan-zededa and @giggsoff for suggestions, comments if any

shjala:
I will fix the yetus complains.

tests/sec/sec_test.go
```diff
@@ -0,0 +1,110 @@
+package sec
```
uncleDecart, line 1: That's a reasonable complaint

milan-zededa (commented):
Overall this is OK, although there is quite a lot of boiler-plate out of which only one line is the actual test. Currently, I would not say that writing eden tests in golang is easier than with escript, but hopefully over time we can improve the test framework and make it more user friendly.

PR #919, opened 2023-11-06
security : check /proc is mounted with secure options

shjala:


shjala:
@uncleDecart a bunch of these tests are going to fail until we fix eve, should we disable sec test suit or it is OK if the fail?

uncleDecart:
>@uncleDecart a bunch of these tests are going to fail until we fix eve, should we disable sec test suit or it is OK if the fail?

@shjala I think it's better to fix tests on EVE. We can rebase EVE test version after that.

PR #928, opened 2023-11-24
Fix nginx test

milan-zededa:
Use `awk` instead of `cut` to make the test robust against any extra space that may occur in the output from `eden pod ps`.
Also fixed is the misspelling in the test file name.


milan-zededa:
Fix for: https://github.com/lf-edge/eden/actions/runs/6811998752/job/18780924635?pr=909#step:3:7088

uncleDecart (approved):

PR #942, opened 2024-01-22
match EVE version for GH actions to the default one

europaul:


europaul:
@uncleDecart some jobs are failing at strange places like starting eden or onboarding EVE. I would restart them, but I don't have the rights. Do you? Or is normal and we can merge?

uncleDecart:
Storage test is known issue, user app test suite is failing on master as well. I'll create respective issues, this could be merged.

uncleDecart (approved):

PR #949, opened 2024-02-05
GitHub Actions: use GitHub Artifact as EVE image if specified

uncleDecart:
Previously, EVE images used for tests in PR were pushed to dockerhub. This comes with several disadvantages, amongst them is dockerhub rate limit and intermediate images published. Now we use GitHub artifacts instead. That still gives you flexibility of downloading specific EVE images to test locally, but removes dockerhub rate limit

Now action user can specify run id and artifact name which contains EVE image, which could be pulled and used to run tests

### DIFF
--- a/.github/actions/run-eden-test/action.yml
+++ b/.github/actions/run-eden-test/action.yml
@@ -15,6 +15,8 @@ inputs:
     type: string
   eve_artifact_name:
     type: string
+  artifact_run_id:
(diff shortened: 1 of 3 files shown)


uncleDecart:
I guess yetus is broken: I dug up the error, hopefully it'll pass now

deitch (approved):

PR #954, opened 2024-02-07
Add support for eden_version input in test workflow

europaul:
When the workflow is triggered from the outside, action/checkout defaults to checking out the latest master which is not always desired. For better compatibility with older EVE release the user should be able to specify the version of eden to be used in the test workflow. If no version is specified, the workflow should resert to the default behavior.

### DIFF
--- a/.github/workflows/eden_emh_arm.yml
+++ b/.github/workflows/eden_emh_arm.yml
@@ -21,7 +21,7 @@ jobs:
           sudo add-apt-repository ppa:longsleep/golang-backports
           sudo apt install -y golang jq expect
       - name: Get eden
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4.1.1
       - name: Create ubuntu server on packet hosting
         id: ubuntu
         run: |
--- a/.github/workflows/eden_gcp.yml
+++ b/.github/workflows/eden_gcp.yml
@@ -30,7 +30,7 @@ jobs:
         fs: ["zfs", "ext4"]
     steps:
       - name: get eden
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4.1.1
       - name: setup go
         uses: actions/setup-go@v3
         with:
--- a/.github/workflows/eden_setup.yml
+++ b/.github/workflows/eden_setup.yml
@@ -16,7 +16,7 @@ jobs:
         run: |
           ip a
       - name: get eden
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4.1.1
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
       - name: Set up QEMU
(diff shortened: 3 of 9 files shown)


europaul:
@uncleDecart this is a WIP, I want to run the tests and see if this implementation works. For the cause of this change see the discussion with @eriknordmark [here](https://github.com/lf-edge/eve/pull/3733)

europaul:
> @europaul Ideally, when referencing workflow in EVE [here](https://github.com/lf-edge/eve/blob/3270a0e52da06decd23184c20df576280579a449/.github/workflows/eden.yml#L69) it should reference and use eden files specified in that branch. There was a bug in checkout I thought I fixed it [here](https://github.com/lf-edge/eden/pull/924) 

@uncleDecart you are right! But I assume we still haven't found a solution how to backport the fix to 0.9.2 - this is why the jobs are failing. How about creating a branch from 0.9.2, perhaps calling it something similar, backporting the commit with the fix to that branch and then referencing the branch in EVE's 11.0-stable

uncleDecart:
make sense

eriknordmark:
One yetus failure:

europaul:
@uncleDecart it looks like the solution from [here](https://github.com/lf-edge/eden/pull/924) doesn't work. See latest comments on the [issue](https://github.com/actions/checkout/issues/1418)  and our own [test results](https://github.com/lf-edge/eve/actions/runs/7816876366/job/21323457573?pr=3733#step:2:470)

Should we use the one proposed in the current PR instead?

eriknordmark (approved):
LGTM

uncleDecart:
Looks good to me, but right now we can't see if it's working or not. I'll create small example to check it out and let you know

uncleDecart:
@europaul [works](https://github.com/uncleDecart/gha-playground/actions/workflows/checking_out.yml). I'd suggest we bump checkout to v4, since there's no reason to use 3.5.3 anymore and let's merge this and backport to 0.9.2-stable branch so that we can point EVE 11-stable to 0.9.2-stable Eden (in case we need any other backporting)

europaul:
> @europaul [works](https://github.com/uncleDecart/gha-playground/actions/workflows/checking_out.yml). I'd suggest we bump checkout to v4, since there's no reason to use 3.5.3 anymore and let's merge this and backport to 0.9.2-stable branch so that we can point EVE 11-stable to 0.9.2-stable Eden (in case we need any other backporting)

I think we can leave the checkout version as it is - we still use v3 for all the other workflows for example and I don't see any new features in the latest release that we would need.

uncleDecart:
@europaul updating to v4 makes sense, because they update to node20, since node16 EOL is 11 sept. 23

uncleDecart (approved):
LGTM, will wait for tests and then merge

PR #962, opened 2024-02-27
Don't delete images dist while clean

uncleDecart:
In case if someone is trying to use local EVE it could lead to repo deletion

### DIFF
--- a/pkg/eden/eden.go
+++ b/pkg/eden/eden.go
@@ -851,11 +851,6 @@ func CleanContext(eveDist, certsDist, imagesDist, evePID, eveUUID, sdnPID, vmNam
 			return fmt.Errorf("CleanContext: error in %s delete: %s", certsDist, err)
 		}
 	}
-	if _, err = os.Stat(imagesDist); !os.IsNotExist(err) {
-		if err = os.RemoveAll(imagesDist); err != nil {
-			return fmt.Errorf("CleanContext: error in %s delete: %s", imagesDist, err)
-		}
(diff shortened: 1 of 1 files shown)


giggsoff:
In that case we will not able to clean current context completely. `./eden clean --current-context` will leave the directory image unconditionally.
Can we use something more precisely, for e.g. delete `dist/CONTEXT-images` (assuming that user-pointed image is outside of dist) or use some flag (to keep image)? Probably the easiest option is to check if `imagesDist` is sub-directory of `eveDist` and if not, keep.

OhmSpectator:
This code has just completely removed my ~/projects, with all the projects inside (tens of them). I demand this code to be eliminated!!! (Well, if the code is fixed and no one can get the same, I'm also happy))

uncleDecart:
@giggsoff , when you want to run EVE locally and you point to an image (or a folder in our case, because I didn't remember) and then you do make clean it'll delete the directory, quite cruel behaviour. Maybe we can delete it only if path to EVE is not specified

giggsoff:
> @giggsoff , when you want to run EVE locally and you point to an image (or a folder in our case, because I didn't remember) and then you do make clean it'll delete the directory, quite cruel behaviour. Maybe we can delete it only if path to EVE is not specified

Current logic is to fill options by default, including EVE path. It is not easy to check if user override it manually.
As I mentioned, we can try to delete `dist/CONTEXT-images` as a path where eden-generated images stored.

Issue #951, opened 2024-02-06
Storage test suite flakiness


uncleDecart:
This is a tracker to collect all Storage test suite runs, which failed in GitHub Actions

### How to contribute?

If you see any failure in GitHub actions runs, please add it to this table. 
#### How to find eden version?
Open workflow file of the run and see which eden tag is invoked, like [here](https://github.com/lf-edge/eve/actions/runs/7802284865/workflow#L51) it is 0.9.3. 
#### How to find EVE version?
In setup-job of workflow which failed find eve image parameter which in inputs. For example, [here](https://github.com/lf-edge/eve/actions/runs/7789442745/job/21247727145#step:1:27) it is evebuild/pr:3698. In case if it contains evebuild/pr put PR in it. 

| Link to run | Eden version | EVE version | test | file system |
|-|-|-|-|-|
|[1](https://github.com/lf-edge/eve/actions/runs/7799540369/job/21278601141?pr=3698#step:3:2460) | 0.9.5 |PR|        `volume_test` | zfs |
|[2](https://github.com/lf-edge/eve/actions/runs/7805364625/job/21292407007#step:3:2512)| 0.9.5 | master | `volumes_test` | zfs |
|[3](https://github.com/lf-edge/eden/actions/runs/7844554055/job/21407948409?pr=961#step:3:2524) | master | 11.7.0 | `volumes_test` | zfs |

### How to run suite locally
Clean previous build and add default config with debug level of printing
```bash
make clean && make build-tests
./eden config add default
./eden config set default --key eve.log-level --value debug
```
(optional) set EVE tag to a version you're testing
```
./eden config set default --key eve.tag --value 11.3.0
```
Setup and run a test
```
./eden setup
./dist/bin/eden+ports.sh 2223:2223
./eden start
./eden eve onboard
./eden test ./tests/workflow -s storage.tests.txt -v debug
```

eriknordmark:
A recent run https://github.com/lf-edge/eve/actions/runs/7806282543/job/21310964084?pr=3742 shows the ZFS storage test suite failing to deploy an app instance due to insufficient memory.
Can we easily  increase the memory for the virtual device?

uncleDecart:
> A recent run https://github.com/lf-edge/eve/actions/runs/7806282543/job/21310964084?pr=3742 shows the ZFS storage test suite failing to deploy an app instance due to insufficient memory.
Can we easily increase the memory for the virtual device?

[...]

PR #963, opened 2024-02-28
Increase default VirtMemory

uncleDecart:
We saw failures in tests, due to insufficient memory, This should solve #951

Increasing in defaults makes more sense rather than in GitHub actions, so that error won't happen locally

cc: @giggsoff @eriknordmark


uncleDecart:
Let's see if the tests will not hit it. Increased mem twice

eriknordmark (approved):
Doesn't seem to make things worse - and presumably better.
See some TOOMANY for nginx etc in the remaining failures which is unrelated.

PR #966, opened 2024-04-19
Fix firmware path for Eden-SDN and define DeviceName

milan-zededa:
[Recent changes](https://github.com/lf-edge/eden/pull/946) broke building of the Eden-SDN VM by moving the firmware files one directory deeper in the dist dir layout.

Another change in this commit is to set `DeviceName` inside the `EdgeDevConfig`. [The Kubevirt-variant of EVE will depend on this](https://github.com/naiming-zededa/eve/blob/poc-dec3/pkg/kube/cluster-init.sh#L114-L136) being set in the received config. However, it does not matter what the value is as long as it is not empty. In Eden will therefore simply use the device UUID as the device name.

### DIFF
--- a/pkg/controller/device.go
+++ b/pkg/controller/device.go
@@ -638,6 +638,7 @@ volumeLoop:
 			Uuid:    dev.GetID().String(),
 			Version: strconv.Itoa(dev.GetConfigVersion()),
 		},
+		DeviceName:         dev.GetID().String(),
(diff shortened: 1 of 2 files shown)


giggsoff (approved):

europaul (approved):
Thank you for fixing this @milan-zededa!

PR #967, opened 2024-04-19
 Test upgrade from LTS to the version from PR, not vice-versa

milan-zededa:
EVE downgrade is generally not supported and it will fail between some versions, see for example:
https://github.com/lf-edge/eve/pull/3812#issuecomment-1999741583

We should instead test upgrade from some LTS version to the latest version with the changes from the tested PR.

@uncleDecart We need to test EVE upgrades and not downgrades. In this PR I'm trying to figure out how to this but github actions are a mystery to me :)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -157,7 +157,7 @@ jobs:
           file_system: ${{ matrix.file_system }}
           tpm_enabled: true
           suite: "eve-upgrade.tests.txt"
-          eve_image: lf-edge/eve:11.0.5-lts
+          eve_image: lfedge/eve:11.0.5-lts
           env_vars: EVE_VERSION=${{ inputs.eve_image }}
           eve_artifact_name: ${{ inputs.eve_artifact_name }}
           artifact_run_id: ${{ inputs.artifact_run_id }}


.github/workflows/test.yml
```diff
@@ -157,7 +157,8 @@ jobs:
           file_system: ${{ matrix.file_system }}
           tpm_enabled: true
           suite: "eve-upgrade.tests.txt"
-          eve_image: ${{ inputs.eve_image }}
+          eve_image: lfedge/eve:11.0.5-lts
+          env_vars: EVE_VERSION=${{ inputs.eve_image }}
```
giggsoff, line 161: As I remember, we expect only something like `11.0.5-lts` as `EVE_VERSION` variable in [test](https://github.com/lf-edge/eden/blob/4fbc73cb9e93fc51416b54662b739ffc2f98f98a/tests/update_eve_image/testdata/update_eve_image_http.txt#L2). At the same time registry may be not aligned with define in test. So it looks like additional adjustment in test required.

.github/workflows/test.yml
```diff
@@ -157,7 +157,8 @@ jobs:
           file_system: ${{ matrix.file_system }}
           tpm_enabled: true
           suite: "eve-upgrade.tests.txt"
-          eve_image: ${{ inputs.eve_image }}
+          eve_image: lfedge/eve:11.0.5-lts
```
giggsoff, line 160: Inside the test [workflow](https://github.com/lf-edge/eden/blob/master/tests/workflow/eve-upgrade.tests.txt) we run several steps:

1. Run and onboard EVE-OS
2. Run update of baseOS image from http store
3. Downgrade to the original baseOS
4. Run update of baseOS image from oci store
5. Downgrade to the original baseOS

In steps 2 and 4 we use different versions of baseOS to not allow EVE-OS just to check another partition and boot from it without downloading and writing. Please do not forget it (for example we can split oci and http different fresh workers).
We use downgrade as we still want to check an ability of current (testing) version to be updated.

giggsoff (commented):
Do we still want to check that current version can handle upgrades to another versions properly? Is it still possible to downgrade to some recent version that we can change from time to time (having part of tests red for short time, but for some clear reason of no compatibility)?

PR #969, opened 2024-05-07
Use latest EVE LTS version for testing upgrades

milan-zededa:
If EVE 12.X is installed, downgrade to 11.X (or older) version is going to fail. This is because the `policy_version` of fscrypt was bumped from 1 to 2 and this change is not backward compatible. See:

https://github.com/lf-edge/eve/commit/8b8424668cc76098f3326c790333502ff5fc51a3

### DIFF
--- a/tests/update_eve_image/testdata/update_eve_image_oci.txt
+++ b/tests/update_eve_image/testdata/update_eve_image_oci.txt
@@ -1,5 +1,5 @@
 # Default EVE version to update
-{{$eve_ver := "9.6.0"}}
+{{$eve_ver := "12.0.1"}}
 
 # Default EVE registry to update
 {{$eve_registry := "lfedge/eve"}}
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -53,7 +53,7 @@ const (
 	DefaultRegistryPort         = 5050
 
 	//tags, versions, repos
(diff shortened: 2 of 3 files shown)


uncleDecart (approved):

tests/update_eve_image/testdata/update_eve_image_oci.txt
```diff
@@ -1,5 +1,5 @@
 # Default EVE version to update
-{{$eve_ver := "9.6.0"}}
+{{$eve_ver := "12.0.1"}}
```
giggsoff, line 2: Can we use different versions for oci and http tests if it is possible? I wrote few words about the reason [here](https://github.com/lf-edge/eden/pull/967#discussion_r1572855910).

milan-zededa: We will have to wait for another 12.X release. Currently there are only two (12.0.0 is broken) and we need one as the base version.

milan-zededa: OK, converted to draft.

eriknordmark: @milan-zededa @uncleDecart We do have a 12.1.0 and 12.2.1 which can be used for this if you just want different versions.
(It would be nice to get less unneeded failures in the eden runs.)

milan-zededa:
Superseded by https://github.com/lf-edge/eden/pull/986

PR #984, opened 2024-06-18
GitHub Actions: Integrate telemetry action

uncleDecart:
This action will gather statistics around eden test execution in GitHub Runners, helping us identify bottlenecks

CC: @milan-zededa 

Let's see what we have in Eden repo

### DIFF
--- a/.github/actions/run-eden-test/action.yml
+++ b/.github/actions/run-eden-test/action.yml
@@ -21,6 +21,11 @@ inputs:
 runs:
   using: 'composite'
   steps:
+    - name: Collect Workflow Telemetry
+      uses: catchpoint/workflow-telemetry-action@v2
+      with:
+        proc_trace_sys_enable: true
+        comment_on_pr: false
     - name: Setup Environment
       uses: ./eden/.github/actions/setup-environment
       with:


milan-zededa (approved):
If this works well we could keep this beyond the current issue troubleshooting.

uncleDecart:
> If this works well we could keep this beyond the current issue troubleshooting.

Yep, I hope it works during failure as well, they should give some outputs :D

milan-zededa:
I see error `Error: [Workflow Telemetry] Resource not accessible by integration', although diagrams are provided...

uncleDecart:
image
image

Huh, CPU load goes in the beginning high, we are not using much resources

uncleDecart:
Seems to be [working](https://github.com/lf-edge/eden/actions/runs/9562869123), adding sys calls information to get better visibility.

giggsoff:
> Seems to be [working](https://github.com/lf-edge/eden/actions/runs/9562869123), adding sys calls information to get better visibility.

Can you please check `Step Trace` section? Looks like it contains only one (or two) steps. Are you sure that other charts contains the whole workflow?

uncleDecart:
you're absolutely right @giggsoff, let's try this one :D

uncleDecart:
Seems like it [works](https://github.com/lf-edge/eden/actions/runs/9564137138?pr=984) time wise, still this steps are weirdly seen, could be because of reusable actions?

uncleDecart:
So am I being punished for ~~over-engineering~~ separating things? Oh man. Well, we can create an action without this hierarchy, but still biggest question is how can we see nested virtualisation impact? It should be something like count of context change, but for that we technically need to get to host, from guest machine maybe sleep time could help?

milan-zededa:
> So am I being punished for ~over-engineering~ separating things? Oh man. Well, we can create an action without this hierarchy, but still biggest question is how can we see nested virtualisation impact? It should be something like count of context change, but for that we technically need to get to host, from guest machine maybe sleep time could help?

Maybe this could be relevant for us: https://scoutapm.com/blog/understanding-cpu-steal-time-when-should-you-be-worried

uncleDecart:
@milan-zededa I think we should merge this and then in my separate fork I'll run this on self-hosted runner so that we can compare numbers and talk to buildjet about it

milan-zededa (approved):

PR #987, opened 2024-06-25
qemu: replace serial console port

christoph-zededa:
7777 is used by gh action runner on IPv6

zededa@eve-runner:~/actions-runner$ sudo netstat -ltnp Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp        0      0 127.0.0.54:53           0.0.0.0:*               LISTEN      1165/systemd-resolv
tcp        0      0 127.0.0.53:53           0.0.0.0:*               LISTEN      1165/systemd-resolv
tcp6       0      0 ::1:7777                :::*                    LISTEN      14603/node
tcp6       0      0 :::22                   :::*
LISTEN      1/init

also some implementations of telnet try IPv6 first if the host has both addresses assigned

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -42,7 +42,7 @@ const (
 	DefaultIP                   = "192.168.0.1"
 	DefaultEVEIP                = "192.168.1.2"
 	DefaultEserverPort          = 8888
-	DefaultTelnetPort           = 7777
+	DefaultTelnetPort           = 17777
 	DefaultQemuMonitorPort      = 7788
 	DefaultQemuNetdevSocketPort = 7790
 	DefaultSSHPort              = 2222
--- a/shell-scripts/collect-info-console.sh
+++ b/shell-scripts/collect-info-console.sh
@@ -18,7 +18,7 @@ for i in $(seq 3); do
   {
     echo "eve verbose off"; echo "eve enter debug"; sleep 3;
     echo "which collect-info.sh"; sleep 3
-  } | telnet localhost 7777 | tee telnet.stdout
+  } | telnet 127.1 17777 | tee telnet.stdout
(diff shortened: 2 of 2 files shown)


uncleDecart (approved):
Thank you @christoph-zededa ! Good catch

uncleDecart:
Maybe it makes sense to actually check if following port is busy and fail, wdyt @christoph-zededa ?

christoph-zededa:
> Maybe it makes sense to actually check if following port is busy and fail, wdyt @christoph-zededa ?

It will fail if a component tries to listen on the same port with the same IP version.

You mean a check that goes through all listeners and checks if the pid of a listener on port x on IPv4 is different than port x on IPv6?

giggsoff:
> Maybe it makes sense to actually check if following port is busy and fail, wdyt @christoph-zededa ?

Right now Eden will fail on step of `eden start` (during start of qemu and waiting for 5 seconds after start in `RunCommandNohup`) if it cannot use port defined in config.

milan-zededa:
@christoph-zededa What about instead adding `eden config set default --key eve.telnet-port --value 17777` into the setup-environment action for your runner? (and then improving `collect-info-console.sh` to read port from the config)
Or is the process `node` that listens on 7777 something that github will run on every self-hosted runner (and thus changing the default makes sense)?

christoph-zededa:
> @christoph-zededa What about instead adding `eden config set default --key eve.telnet-port --value 17777` into the setup-environment action for your runner? (and then improving `collect-info-console.sh` to read port from the config) Or is the process `node` that listens on 7777 something that github will run on every self-hosted runner (and thus changing the default makes sense)?

As far as I understand these github runners, they have to run on every self-hosted runner.

milan-zededa (approved):

PR #988, opened 2024-06-26
GitHub Actions: Login to DockerHub

uncleDecart:
Testing hits DockerHub pull rate limit. Previosly considered migration option ghcr, has pull rate limits now as well. Step to solving this problem is to add logging in, that way it'll be 200 pulls per day instead of 100 per shared IP address with BuildJet runners


uncleDecart:
PR to merge into my own branch [here](https://github.com/lf-edge/eden/pull/989)

giggsoff:
Is it secure to pass secrets via inputs?

uncleDecart:
> Is it secure to pass secrets via inputs?

Yep @giggsoff , we do that in EVE repository for run-make action [here](https://github.com/lf-edge/eve/blob/master/.github/workflows/publish.yml#L116C33-L116C63)

giggsoff:
> > Is it secure to pass secrets via inputs?
> 
> Yep @giggsoff , we do that in EVE repository for run-make action [here](https://github.com/lf-edge/eve/blob/master/.github/workflows/publish.yml#L116C33-L116C63)

I can see special [option](https://docs.github.com/en/actions/using-workflows/reusing-workflows#passing-inputs-and-secrets-to-a-reusable-workflow) to pass secrets to reusable workflows, is it better to use it instead?

milan-zededa (approved):

giggsoff:
Looks like it still has problems: [link](https://github.com/lf-edge/eden/actions/runs/9683048618). Or am I missing something?

uncleDecart:
Okay, looks like it's fixed

.github/workflows/eden.yml
```diff
@@ -16,4 +16,4 @@ jobs:
     uses: ./.github/workflows/test.yml
     with:
       eve_image: "lfedge/eve:12.4.0"
-
+    secrets: inherit
```
giggsoff, line 19: As I understand from GH docs (I shared the link somewhere in discussions) in that case you should point onto secrets in nested workflows, not onto inputs.

uncleDecart: In this context test.yml is a nested workflow to eden.yml

giggsoff: I mean you inherit secrets (no pass inputs) now. But try to use inputs in nested workflows (not secrets).

uncleDecart: Inside test.yml we use them as inputs like [here](https://github.com/lf-edge/eden/pull/988/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R79) test.yml is invoked by EVE repository as reusable workflow, so we should inherit secrets for test.yml, eden.yml does the same. If we are to add inputs to test.yml it'll be as it was before in EVE repository

uncleDecart: Any ideas how to shut yetus up on this secret keyword?

.github/workflows/test.yml
```diff
@@ -14,6 +14,14 @@ on:  # yamllint disable-line rule:truthy
...
       artifact_run_id:
         type: string
+      docker_account:  # if not provided: use anonymous docker user
+        type: string
+        required: false
+        default: ''
+      docker_token:
+        type: string
+        required: false
+        default: ''
```
giggsoff, line 24: Where do we use these lines?

.github/workflows/eden.yml
giggsoff: Let's try to create `.yetus/detsecrets-ignored-hashes.txt`  with `3e26d6750975d678acb8fa35a0f69237881576b0` line inside.

PR #991, opened 2024-06-27
Test legacy buildjet runner

milan-zededa:
Let's see if Virtualization test suite passes...


uncleDecart (approved):

milan-zededa:
I will rerun Virtualization test suite but it is not looking good...

milan-zededa:
@uncleDecart Sometimes the Virtualization test suite passes and other times it fails right away. I think it is not related to runner version but rather on which server in the buildjet pool it is scheduled. Some of those servers are probably misconfigured.

PR #993, opened 2024-06-27
Increase timeout for deployment of two nginx apps

milan-zededa:
10 minutes is sometimes not enough with ZFS and slower runner.

### DIFF
--- a/tests/app/testdata/2dockers_test.txt
+++ b/tests/app/testdata/2dockers_test.txt
@@ -17,7 +17,7 @@ eden -t 1m pod deploy -p 8028:80 docker://nginx -n t2 --memory 512MB
 stdout 'deploy pod t2 with docker://nginx request sent'
 
 # Wait for run
-test eden.app.test -test.v -timewait 10m RUNNING t1 t2
+test eden.app.test -test.v -timewait 15m RUNNING t1 t2
 
 # Dockers detecting
 eden -t 1m pod ps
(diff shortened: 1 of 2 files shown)


uncleDecart (approved):
LGTM, but I'm wondering if this is hacking our way through async programming? Or another words, can we write code so that it'll be all on async handlers and there will be no need for timeouts

milan-zededa:
> LGTM, but I'm wondering if this is hacking our way through async programming? Or another words, can we write code so that it'll be all on async handlers and there will be no need for timeouts

This is not sleep. We deploy app than wait for it to be RUNNING. As soon as we receive info that it is running, test continues.
But there has to be some timeout to eventually give up if eve is failing to deploy.

uncleDecart:
> This is not sleep. We deploy app then wait for it to be RUNNING. As soon as we receive info that it is running, test continues.
But there has to be some timeout to eventually give up if eve is failing to deploy.

Do you think it makes sense to have timeouts for each operation or should it be one timeout for whole test suite rather than per test?

milan-zededa:
> > This is not sleep. We deploy app then wait for it to be RUNNING. As soon as we receive info that it is running, test continues.
> > But there has to be some timeout to eventually give up if eve is failing to deploy.
> 
> Do you think it makes sense to have timeouts for each operation or should it be one timeout for whole test suite rather than per test?

For the whole test suite we would have to use pretty large timeout. I don't think it would make sense to wait for an hour for e.g. an app to start running.

uncleDecart:
Makes sense

PR #997, opened 2024-07-04
Try to re-enable accel

milan-zededa:
Now that runners are supposedly fixed, let's see how all tests fare with acceleration enabled...


.github/actions/run-eden-test/action.yml
```diff
@@ -55,6 +55,9 @@ runs:
       run: EDEN_TEST_STOP=n ./eden test ./tests/workflow -s ${{ inputs.suite }} -v debug
       shell: bash
       working-directory: "./eden"
+    - name: Setup tmate session
+      if: ${{ failure() }}
+      uses: mxschmitt/action-tmate@v3
```
giggsoff, line 60: We have docker login step now (was it enabled?), assume it stores token (or password) on VM unencrypted as usual. Be careful with tmate sessions.

milan-zededa: Good point, I can revert commit with docker login for my testing.
Although I'm not sure if those secrets are actually set in the eden repo (they for sure are in the eve repo).

giggsoff: It is quite difficult to go through output of reusable workflow... In any case, you need to keep in mind a possible problem.

milan-zededa:
Closing since buildjet failed to provide a fix.

PR #999, opened 2024-07-10
GitHub Actions use org secrets

deitch:
lfedge created a new user for all of the workflows, and saved the credentials as org secrets, specifically `RELEASE_DOCKERHUB_ACCOUNT` and `RELEASE_DOCKERHUB_TOKEN`. This PR removes the different name in the flows and uses the org standard `secrets.RELEASE_DOCKERHUB_ACCOUNT` and `secrets.RELEASE_DOCKERHUB_TOKEN`.

Eden actually was using both, `RELEASE_DOCKERHUB_*` and `DOCKERHUB_*`, with only a few left using the `DOCKERHUB_*`. This unifies it all.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -17,7 +17,7 @@ jobs:
         if: ${{ env.DOCKERHUB_TOKEN != '' }}
         run: echo "defined=true" >> $GITHUB_OUTPUT
         env:
-          DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+          DOCKERHUB_TOKEN: ${{ secrets.RELEASE_DOCKERHUB_TOKEN }}
   build:
     runs-on: ubuntu-22.04
     needs: [check-secrets]
@@ -32,8 +32,8 @@ jobs:
       - name: Login to DockerHUB
         id: login
         run: |
-          echo "${{ secrets.DOCKERHUB_TOKEN }}" |\
-             docker login -u "${{ secrets.DOCKERHUB_USERNAME }}" --password-stdin
+          echo "${{ secrets.RELEASE_DOCKERHUB_TOKEN }}" |\
+             docker login -u "${{ secrets.RELEASE_DOCKERHUB_ACCOUNT }}" --password-stdin
       - name: Build and push
         run: |
           make DOCKER_TARGET=push LINUXKIT_TARGET=push DOCKER_PLATFORM=linux/arm64,linux/amd64 build-docker


yash-zededa (approved):

milan-zededa:
Note that for tests running for an open PR this dockerhub credentials will not be used unless we [switch pull_request to pull_request_target](https://github.com/lf-edge/eden/blob/master/.github/workflows/eden.yml#L4). But I think that in this repo it is not so important (we do not open PRs that often here) and it has security risks.
I'm not sure about [pull_request_review that we use in the eve repo](https://github.com/lf-edge/eve/blob/master/.github/workflows/eden.yml#L14) - does it have access to secrets?

milan-zededa (approved):

milan-zededa:
However, it seems that in each test workflow the only non-OSS pull is for the nginx container. Hence it is not so import to use the dockerhub account for eden tests.

deitch:
> However, it seems that in each test workflow the only non-OSS pull is for the nginx container (or is it OSS, how can I check?).

nginx is not OSS. You look at the repo. For example, [lfedge/eve](https://hub.docker.com/r/lfedge/eve) has this:

Screenshot 2024-07-10 at 11 38 34 AM

While nginx does not:

Screenshot 2024-07-10 at 11 39 07 AM

"Docker official images" explicitly are subject to pull limits.

> Hence it is not so import to use the dockerhub account for eden tests.

So maybe we do not need it at all? Perhaps. Let's this one in, which just switches the secret used to unify it. Then we can remove secret access entirely in a separate one if we want.

milan-zededa:
> So maybe we do not need it at all? Perhaps. Let's this one in, which just switches the secret used to unify it. Then we can remove secret access entirely in a separate one if we want.

Note that I'm not against using docker login in tests. It is just that these secrets are not available to an open PR (unless we use pull_request_target) and thus we will continue running them with anonymous docker user.

uncleDecart:
Eden was using inherited secrets from EVE, hence the naming release. We used secrets instead of inputs as per GitHub guidelines. I wonder though if any repository using reusable Eden workflow will have access to this credentials

milan-zededa:
> Eden was using inherited secrets from EVE, hence the naming release. We used secrets instead of inputs as per GitHub guidelines. I wonder though if any repository using reusable Eden workflow will have access to this credentials

The real problem is here: https://github.com/lf-edge/eve/blob/master/.github/workflows/eden.yml#L14
`pull_request_review` does not grant access to secrets, hence no secrets are propagated to eden (those fields are inherited empty).

PR #1000, opened 2024-07-15
GitHub Actions: Bump reusable actions to be compatible with node16

uncleDecart:
Following actions were using node12 which is deprecated and they were forced to use node16. Also upload-release-asset was deprecated so this patch switches to maintained action listed in official repository

### DIFF
--- a/.github/workflows/eden_setup.yml
+++ b/.github/workflows/eden_setup.yml
@@ -18,9 +18,9 @@ jobs:
       - name: get eden
         uses: actions/checkout@v4.1.1
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
+        uses: docker/setup-buildx-action@v3
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
+        uses: docker/setup-qemu-action@v3
       - name: setup go
         uses: actions/setup-go@v3
         with:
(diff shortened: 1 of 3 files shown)


uncleDecart:
cc @yash-zededa for some reason I can't add you to reviewers

uncleDecart:
I believe this is the reason why this workflow broke https://github.com/lf-edge/eden/actions/runs/9829563875 (last week I released eden 0.10.0)

yash-zededa:
> cc @yash-zededa for some reason I can't add you to reviewers

I can't add my review. It looks like I need to be a collaborator for the Eden repo. @uncleDecart, do you have access to add me as a collaborator?

uncleDecart:
I have no power here, we have to ask @deitch or @eriknordmark for that

eriknordmark (approved):
LGTM

giggsoff:
I can see that GH expect new versions for another actions as well ([here](https://github.com/lf-edge/eden/actions/runs/9938274553) in annotations):
```
The following actions uses Node.js version which is deprecated and will be forced to run on node20: actions/setup-go@v3, actions/upload-artifact@v3. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/
```
Do you plan to update them separately?

uncleDecart:
Yes @giggsoff , I'll do them separately :)

PR #1004, opened 2024-07-17
eden/qemu: start QMP logger to log QEMU events

rouming:
Nice to have QMP events from QEMU, probably can help to debug sudden reboots/reset issues.

This commit uses simple shell "echo | socat" cmd for reading qemu qmp events. Probably makes sense to rewrite on native go.

QEMU creates a socket file and uses is as server socket. socat attaches to it, sends handshake message and redirects output (qmp events) to the `qmp.log` file. if socket file was not created within 5 seconds, the `startQMPLogger()` complains, but this error is not considered as fatal and eden continues.

### DIFF
--- a/.github/actions/publish-logs/action.yml
+++ b/.github/actions/publish-logs/action.yml
@@ -17,6 +17,7 @@ runs:
         ./eden metric --format json > metric.log || echo "no metric"
         ./eden netstat --format json > netstat.log || echo "no netstat"
         cp dist/default-eve.log console.log || echo "no device log"
+        cp dist/qmp.log qmp.log || echo "no qmp log"
         docker logs eden_adam > adam.log 2>&1 || echo "no adam log"
       shell: bash
       working-directory: "./eden"
@@ -53,4 +54,5 @@ runs:
             ${{ github.workspace }}/eden/metric.log
             ${{ github.workspace }}/eden/netstat.log
             ${{ github.workspace }}/eden/console.log
+            ${{ github.workspace }}/eden/qmp.log
             ${{ github.workspace }}/eden/adam.log
--- a/pkg/eden/qemu.go
+++ b/pkg/eden/qemu.go
@@ -11,6 +11,7 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/lf-edge/eden/pkg/defaults"
 	"github.com/lf-edge/eden/pkg/edensdn"
@@ -39,6 +40,34 @@ func StopSWTPM(stateDir string) error {
 	command := "swtpm"
 	pidFile := filepath.Join(stateDir, fmt.Sprintf("%s.pid", command))
 	return utils.StopCommandWithPid(pidFile)
+}
+
+func startQMPLogger(qmpSockFile string, qmpLogFile string) error {
+	shellcmd := fmt.Sprintf(
+		"echo '{\"execute\": \"qmp_capabilities\"}' | " +
+		"socat -t0 -,ignoreeof UNIX-CONNECT:%s > %s",
+		qmpSockFile, qmpLogFile)
+	opts := []string{
+		"-c", shellcmd,
+	}
+
+	var err error
+
+	// Retry a few times if socket is not available yet
+	n := 5
+	for n > 0 {
+		if err = utils.RunCommandNohup("sh", "", "", opts...); err != nil {
+			time.Sleep(1 * time.Second)
+			n--
+			continue
+		}
+		break
+	}
+	if err != nil {
+		 return fmt.Errorf("startQMPLogger: can't connect to the QMP socket, presumably QEMU did not start")
+	}
+
+	return nil
 }
 
 // StartEVEQemu function run EVE in qemu
@@ -193,6 +222,12 @@ func StartEVEQemu(qemuARCH, qemuOS, eveImageFile, imageFormat string, isInstalle
 		qemuOptions += fmt.Sprintf("-readconfig %s ", qemuConfigFile)
 	}
 
+	qmpSockFile := filepath.Join(filepath.Dir(pidFile), "qmp.sock")
+	qmpLogFile := filepath.Join(filepath.Dir(pidFile), "qmp.log")
+
+	// QMP sock
+	qemuOptions += fmt.Sprintf("-qmp unix:%s,server,wait=off", qmpSockFile)
+
 	log.Infof("Start EVE: %s %s", qemuCommand, qemuOptions)
 	if foreground {
 		if err := utils.RunCommandForeground(qemuCommand, strings.Fields(qemuOptions)...); err != nil {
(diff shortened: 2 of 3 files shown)


pkg/eden/qemu.go
```diff
@@ -193,6 +225,9 @@ func StartEVEQemu(qemuARCH, qemuOS, eveImageFile, imageFormat string, isInstalle
 		qemuOptions += fmt.Sprintf("-readconfig %s ", qemuConfigFile)
 	}
 
+	// QMP sock
+	qemuOptions += fmt.Sprintf("-qmp unix:%s,server,wait=off", qmpSock)
```
uncleDecart, line 229: And if we can't start this thing, qmpSock will be file that does not exist, will qemu just send warning message or will it crash in that case?

rouming: QEMU creates a socket file and uses is as server socket. socat attaches to it, sends handshake message and redirects output to a file. if socket file was not created for 5 seconds, the `startQMPLogger()` complains, but this error is not considered as fatal.

pkg/eden/qemu.go
```diff
@@ -41,10 +43,40 @@ func StopSWTPM(stateDir string) error {
...
+		return fmt.Errorf("startQMPLogger: no QMP socket")
+	}
+	shellcmd := fmt.Sprintf(
+		"echo '{\"execute\": \"qmp_capabilities\"}' | " +
+		"socat -t0 -,ignoreeof UNIX-CONNECT:%s > %s",
+		qmpSock, qmpLog)
+	opts := []string{
+		"-c", shellcmd,
+	}
+	if err := utils.RunCommandNohup("sh", "", "", opts...); err != nil {
```
milan-zededa, line 69: Will this command stop when qemu exits?
If not, then we will have to somehow stop it from here: https://github.com/lf-edge/eden/blob/master/pkg/eden/qemu.go#L211-L213

rouming: Yes. socat dies when socket is closed.

pkg/eden/qemu.go
```diff
@@ -41,10 +42,39 @@ func StopSWTPM(stateDir string) error {
...
+		"socat -t0 -,ignoreeof UNIX-CONNECT:%s > %s",
+		qmpSock, qmpLog)
+	opts := []string{
+		"-c", shellcmd,
+	}
+
+	var err error
+
+	// Retry a few times if socket is not available yet
+	n := 5
```
milan-zededa, line 57: I suspect that on these runners we may need to give more than 5 seconds for qemu to start.

rouming:
So folks (@uncleDecart @milan-zededa), I need some help. First of all very first commit smoothly works locally, but does not work on GH. The problem which I chase for a couple of days in a row is different content of `currentPath`, for example:
```
	qmpSock := filepath.Join(currentPath, defaults.DefaultDist, "qmp.sock")
```
(quite common construction used everywhere in eden to get some path).
This `qmpSock` has the following path when I run eden locally:
`/home/roman/devel/zededa/eden/dist`, and `dist` folder exists (contains all the files generated by eden) to the point when my code is executed, so I'm able to create `qmp.log` and `qmp.sock` files.

When I run same patch on GH, I see that `qmpSock` is equal to the following path:
`/tmp/go-test-script854418973/script-eden_start/dist`, and `dist` in this particular case does not exist, so I'm not able to create any files in this folder. In actual action GH output I see this paths quite often: 
`/home/runner/actions-runner/_work/eden/eden/eden/dist/default-eve.pid`, so I was expecting, that `currentPath` would be equal to that, but this is not the case.

So I'm super confused and is not clear to me how to build the correct path for some files. Any help is appreciated.

rouming:
Ok, the problem is solved by taking directory of the `evePidFile`, which always exists in the correct directly. I failed to create my own configuration due to the fact that it is always empty, I failed to use the `currentPath` due to the fact, that it is different from actual runner dist directory `/home/runner/actions-runner/_work/eden/eden/eden/dist` (I assume this is because we do `os.Chdir()` somewhere and current work directory becomes different only for GH runners). Anyway, for this minor PR this approach should work and does not look super ugly. Just a little bit. Little bit disgusting and crooked. Little. Bit.

rouming:
For those who care: this line is likely broken and has similar problems I had: 
https://github.com/lf-edge/eden/blob/04badba3bdf1053ae64a3773cf681c46528ab80e/pkg/openevec/eve.go#L148
Same construction of using the `currentPath` and attempt to create a file further in the `utils.CreateUsbNetConfImg()` call.

milan-zededa:
> Ok, the problem is solved by taking directory of the `evePidFile`, which always exists in the correct directly. I failed to create my own configuration due to the fact that it is always empty, I failed to use the `currentPath` due to the fact, that it is different from actual runner dist directory `/home/runner/actions-runner/_work/eden/eden/eden/dist` (I assume this is because we do `os.Chdir()` somewhere and current work directory becomes different only for GH runners). Anyway, for this minor PR this approach should work and does not look super ugly. Just a little bit. Little bit disgusting and crooked. Little. Bit.

This is fine for now. I will also investigate why it is broken. We haven't yet tried onboarding using (fake) USB stick on github runners, which would try to use `usbImagePath. So if it has the same problem it has not been noticed yet.

milan-zededa (approved):

uncleDecart (approved):

uncleDecart:
I'll rerun failed tests and will merge afterwards

uncleDecart:
Failure isn't something to do with your PR, merging it

.github/actions/publish-logs/action.yml
```diff
@@ -17,6 +17,7 @@ runs:
         ./eden metric --format json > metric.log || echo "no metric"
         ./eden netstat --format json > netstat.log || echo "no netstat"
         cp dist/default-eve.log console.log || echo "no device log"
+        cp dist/qmp.log qmp.log || echo "no qmp log"
```
giggsoff, line 20: Sorry, seems I come too late. Small note here: we have `default-` prefix for reason (above). It is config context (we can create another with non-default name) we may use to run several EVE-OS under Eden control (I know, it is not easy because of port's conflict, but at least possible with some crafted config). With two qemu expected to run on the same qmp.sock it become impossible. So it would be great to not use dist/qmp.log and dist/qmp.sock, but use ones prefixed by context name.

rouming: This is easy to fix, I can do this. I expected this kind of thing, and my initial thoughts to have "default-" prefix was correct, but I had another problem with the path, so forgot. Thanks for bringing that back.

If it won't be difficult for you, could you please take a look into my problem with the `currentPath` and `os.Getwd()`, described here: https://github.com/lf-edge/eden/pull/1004#issuecomment-2247917847

rouming: Here https://github.com/lf-edge/eden/pull/1007

giggsoff: Well, inside of tests we use temp directory as a current to keep all files and artifacts there. So you cannot use current directory there to point onto dist files. You can try to use `utils.ResolveAbsPath` instead which will load use stored path to dict directory in config and append the argument to it.

PR #1005, opened 2024-07-24
Fix http datastore fqdn address

shjala:
In case the http datastore address has a port number, the datastore address should contain both the hostname and the port.

### DIFF
--- a/pkg/expect/http.go
+++ b/pkg/expect/http.go
@@ -96,7 +96,7 @@ func (exp *AppExpectation) checkDataStoreHTTP(ds *config.DatastoreConfig) bool {
 		if err != nil {
 			log.Fatal(err)
 		}
-		if exp.httpDirectLoad && ds.Fqdn == fmt.Sprintf("%s://%s", u.Scheme, u.Hostname()) {
+		if exp.httpDirectLoad && ds.Fqdn == fmt.Sprintf("%s://%s", u.Scheme, u.Host) {
 			return true
 		}
 	}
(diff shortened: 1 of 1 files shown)


uncleDecart (approved):
LGTM, thanks @shjala ! Could you please add comment from your commit to the change, so it won't be gone if we rewrite commit? I think it's a useful comment for people :)

shjala:
> LGTM, thanks @shjala ! Could you please add comment from your commit to the change, so it won't be gone if we rewrite commit? I think it's a useful comment for people :)

I'm not sure what you mean, this is the commit message, should I add more to it?
```
Fix http datastore fqdn address
In case the http datastore address has a port number, the datastore address
should contain both the hostname and the port.

Signed-off-by: Shahriyar Jalayeri <shahriyar@zededa.com>
```

shjala:
@uncleDecart got it, added.

pkg/expect/http.go
```diff
@@ -143,6 +143,11 @@ func (exp *AppExpectation) createDataStoreHTTP(id uuid.UUID) *config.DatastoreCo
 			ds.DType = config.DsType_DsHttps
 		}
 		ds.Fqdn = fmt.Sprintf("%s://%s", u.Scheme, u.Hostname())
+		// In case the http datastore address has a port number,
+		// the datastore address should contain both the hostname and the port.
+		if u.Port() != "" {
+			ds.Fqdn = fmt.Sprintf("%s:%s", ds.Fqdn, u.Port())
```
giggsoff, line 149: Can we use `u.Host` instead in both cases (as full host:port notation), looks like it contains the port and `u.Port()` just extracts it from `u.Host`.

giggsoff (commented):
Please align the logic [here](https://github.com/lf-edge/eden/blob/ffbccba605c3f3782084b43a9ee4664e1a18164f/pkg/expect/http.go#L99).

pkg/expect/http.go
shjala: Sure.

giggsoff (approved):
Thank you! Nice catch.

PR #1010, opened 2024-08-08
Run acl test twice - with and without flow logging 

milan-zededa:
Soon EVE will allow to disable flow logging (and it will become the default behavior),
which reduces the amount of data published to the controller and simplifies iptables
rules installed by EVE.
See: https://github.com/lf-edge/eve-api/pull/62

However, the ACL implementation is quite different between enabled and
disabled flow logging. Therefore, it makes sense to run the ACL test twice
to test both cases.

### DIFF
--- a/tests/workflow/networking.tests.txt
+++ b/tests/workflow/networking.tests.txt
@@ -1,5 +1,5 @@
 # Number of tests
-{{$tests := 13}}
+{{$tests := 14}}
 # EDEN_TEST_SETUP env. var. -- "y"(default) performs the EDEN setup steps
 {{$setup := "y"}}
 {{$setup_env := EdenGetEnv "EDEN_TEST_SETUP"}}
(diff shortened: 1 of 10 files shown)


tests/workflow/networking.tests.txt
```diff
@@ -32,25 +32,27 @@ eden.escript.test -test.run TestEdenScripts/eden_onboard
 eden.escript.test -test.run TestEdenScripts/template_check
 {{end}}
 
-/bin/echo Eden basic network test (5/{{$tests}})
-eden.escript.test -testdata ../network/testdata/ -test.run TestEdenScripts/network_test
+#/bin/echo Eden basic network test (5/{{$tests}})
+#eden.escript.test -testdata ../network/testdata/ -test.run TestEdenScripts/network_test
```
giggsoff, line 36: They are not required anymore? Or it is some lines come from your debugging process?

milan-zededa: Good catch! I left it there accidentally.
It is fixed now.

uncleDecart (approved):
LGTM, do we need to bump EVE version as well, since we're updating API? 

Edit: Okay, we can merge once EVE is released and then we will have to release new Eden

milan-zededa:
> LGTM, do we need to bump EVE version as well, since we're updating API?
> 
> Edit: Okay, we can merge once EVE is released and then we will have to release new Eden

No need, test passes also for older EVE.
Actually, I would prefer to first merge this, release eden and bring this to EVE. Otherwise, my next EVE PR will fail networking tests.

PR #1017, opened 2024-08-23
add package evetestkit

shjala:
This is auxiliary package to make testing eve and vms easier, it offers a set of helper functions to create and manage eve and vm instances.

_I moved it out of my other PRs, it would be great if we can get this review first so I can clean up my other tests (vcom, aziot and sec)._

### DIFF
--- a/pkg/evetestkit/utils.go
+++ b/pkg/evetestkit/utils.go
@@ -1,0 +1,499 @@
+package evetestkit
+
+import (
+	"fmt"
+	"io"
+	"net"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/docker/docker/pkg/namesgenerator"
+	"github.com/dustin/go-humanize"
+	"github.com/lf-edge/eden/pkg/controller"
+	"github.com/lf-edge/eden/pkg/defaults"
+	"github.com/lf-edge/eden/pkg/device"
+	"github.com/lf-edge/eden/pkg/eve"
+	"github.com/lf-edge/eden/pkg/openevec"
+	"github.com/lf-edge/eden/pkg/projects"
+	"github.com/lf-edge/eden/pkg/tests"
+	"github.com/lf-edge/eden/pkg/utils"
+	"github.com/tmc/scp"
+	"golang.org/x/crypto/ssh"
+	"golang.org/x/exp/rand"
+)
+
+const (
+	// AppDefaultSSHPass is a default ssh password for the VM running on the EVE node
+	AppDefaultSSHPass = "passw0rd"
+	// AppDefaultSSHUser is a default ssh user for the VM running on the EVE node
+	AppDefaultSSHUser = "ubuntu"
+	// AppDefaultCloudConfig is a default cloud-init configuration for the VM which just
+	// enables ssh password authentication and sets the password to "passw0rd".
+	AppDefaultCloudConfig = "#cloud-config\npassword: " + AppDefaultSSHPass + "\nchpasswd: { expire: False }\nssh_pwauth: True\n"
+)
+
+var (
+	controllerVerbosiry = "warn"
+	edenConfEnv         = defaults.DefaultConfigEnv
+)
+
+// appInstanceConfig is a struct that holds the information about the app
+// running on the EVE node
+type appInstanceConfig struct {
+	name    string
+	sshPort string
+	sshUser string
+	sshPass string
+}
+
+// EveNode is a struct that holds the information about the remote node
+type EveNode struct {
+	controller *openevec.OpenEVEC
+	edgenode   *device.Ctx
+	cfg        *openevec.EdenSetupArgs
+	tc         *projects.TestContext
+	apps       []appInstanceConfig
+	ip         string
+}
+
+// AppOption is a function that sets the configuration for the app running on
+// the EVE node
+type AppOption func(n *EveNode, appName string)
+
+// TestOption is a function that sets the configuration for the test
+type TestOption func()
+
+func getEdenConfig() (*openevec.EdenSetupArgs, error) {
+	conf := os.Getenv(edenConfEnv)
+	configName := utils.GetConfig(conf)
+	cfg, err := openevec.FromViper(configName, controllerVerbosiry)
+	if err != nil {
+		return nil, fmt.Errorf("can't get the config: %w", err)
+	}
+
+	return cfg, nil
+}
+
+func getOpenEVEC() (*openevec.OpenEVEC, *openevec.EdenSetupArgs, error) {
+	cfg, err := getEdenConfig()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return openevec.CreateOpenEVEC(cfg), cfg, nil
+}
+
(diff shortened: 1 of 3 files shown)


pkg/evetestkit/utils.go
```diff
@@ -0,0 +1,475 @@
...
+
+	if _, err := node.controller.PodDelete(appName, true); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// AddApp adds an app to the EVE node list of apps
+func (node *EveNode) AddApp(appName, SSHUser, SSHPass, SSHPort string) {
```
milan-zededa, line 220: What is the purpose of this function? Based on the name I would think that it would deploy a new app but it just adds some metadata to an internal list.

milan-zededa (approved):
I think this is a good start towards a framework for eden Go tests. Over time we will extend it with more methods (deploy network instance, check published info messages, use Eden-SDN, etc.) as we continue adding more Go tests.

pkg/evetestkit/utils.go
shjala: this adds the app to a internal list to free us from the trouble of passing ssh user/pass/port to SSH and SCP functions.

shjala:
```
// SSHExec executes a command on the app VM running on the EVE node,
// app needs to be added using AddApp before calling this function.
func (node *EveNode) SSHExec(appName, command string) (string, error) {
```

shjala: do you have a suggestion for a better name?

milan-zededa: Could this be perhaps [option(s)](https://dev.to/kittipat1413/understanding-the-options-pattern-in-go-390c) for `DeployVm` function (e.g. `WithSSH(username, password, port)`)?

pkg/evetestkit/utils.go
```diff
@@ -0,0 +1,487 @@
...
+		return nil
+	}
+
+	return nil
+}
+
+// WithSSH is an option that sets the SSH configuration for the app running on
+// the EVE node, this should be use with DeployVM function.
+func WithSSH(user, pass, port string) VMOption {
+	return func(n *EveNode) {
```
milan-zededa, line 323: I would suggest to use `*appInstanceConfig` as the function argument to avoid `n.apps[len(n.apps)-1]`

pkg/evetestkit/utils.go
```diff
@@ -0,0 +1,487 @@
...
+		// slices are ordered, so the last element is the one
+		// we're going to set things for.
+		n.apps[len(n.apps)-1].sshUser = user
+		n.apps[len(n.apps)-1].sshPass = pass
+		n.apps[len(n.apps)-1].sshPort = port
+	}
+}
+
+// DeployVm deploys a VM on the EVE node
+func (node *EveNode) DeployVM(imageURL string, pc openevec.PodConfig, options ...VMOption) error {
```
milan-zededa, line 333: Btw. if this is not limited to VMs but can be used for containers, then maybe `DeployApp` is a better name

shjala: I haven't tested containers, but I guess it should work with a container pod config. I rename it.

pkg/evetestkit/utils.go
```diff
@@ -0,0 +1,487 @@
...
+	pc.StartDelay = 0
+	pc.PinCpus = false
+
+	return pc
+}
+
+// SetControllerVerbosity sets the verbosity level of the controller,
+// possible values are: panic, fatal, error, debug, info, trace, warn
+// this should be set before calling InitilizeTest.
+func SetControllerVerbosity(verbosity string) {
```
milan-zededa, line 405: `SetControllerVerbosity` and `SetEdenConfigEnv` could be options for `InitilizeTest`

pkg/evetestkit/utils.go
```diff
@@ -0,0 +1,487 @@
...
+	cfg, err := getEdenConfig()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return openevec.CreateOpenEVEC(cfg), cfg, nil
+}
+
+// CreateEveNode creates a new EveNode struct
+func CreateEveNode(node *device.Ctx, tc *projects.TestContext) (*EveNode, error) {
```
milan-zededa, line 83: Should we keep this exported even though EveNode is returned by `InitilizeTest`?

pkg/evetestkit/utils.go
shjala: what if we want to add other options in the future that need some other kind of argument, I though EveNode  is general enough to extract and set, and in the future we can extend EveNode without affecting WithXXX functions.

pkg/evetestkit/utils.go
```diff
@@ -0,0 +1,487 @@
...
+
+	return node.controller.PodDeploy(imageURL, pc, node.cfg)
+}
+
+// IsTpmEnabled checks if EVE node is running with (SW)TPM enabled
+func (node *EveNode) IsTpmEnabled() bool {
+	return node.cfg.Eve.TPM
+}
+
+func (node *EveNode) setEveIP() error {
```
milan-zededa, line 347: Shouldn't this be `getEveIP` or `discoverEveIP` or `retrieveEveIP` ... (since we are not really changing the Node IP address)

pkg/evetestkit/utils.go
shjala: change it to accept appName, it is harmless enough that furue WithXXX can ignore it.

pkg/evetestkit/utils.go
```diff
@@ -0,0 +1,487 @@
...
+	output, err := session.CombinedOutput(command)
+	if err != nil {
+		return "", fmt.Errorf("failed to run command: %s", err)
+	}
+
+	return string(output), nil
+}
+
+// SCPCopy copies a file from the local machine to the app VM running on the EVE node.
+func (node *EveNode) SCPCopy(appName, localFile, remoteFile string) error {
```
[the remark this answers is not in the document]
shjala: I thought the function signature makes it clear that it is applied to app, but sure.

pkg/evetestkit/utils.go
milan-zededa: OK, I like `getAppConfig`,  `n.apps[len(n.apps)-1]` looked a bit odd

pkg/evetestkit/utils.go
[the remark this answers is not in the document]
milan-zededa: It is clear but we need to allow SSH/SCP/WaitForSSH for EVE itself (can be called `SSHExecEVE`, etc., I don't have a preference).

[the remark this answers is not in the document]
shjala: prefixed them with App and Eve.

milan-zededa (approved):
OK, just couple yetus complains to fix

milan-zededa (approved):

uncleDecart:
I'm late to the party I see :D Thanks @shjala , I think it's a great start, we will shape it as we go and add more tests

PR #1011, opened 2024-08-13
[wip] add aziot test

shjala:

PR #1018, opened 2024-08-26
add secrets for aziot test

shjala:
Aure-Iot-Edge test need access to some secrets to onboard a device into the Azure portal before testing, secrets are already added to the repository but #1011  fails to access them, according to Github documentation "_GitHub Actions can only read a secret if you explicitly include the secret in a workflow._"

Hope the syntax is correct.


shjala:
closing this one, check https://github.com/lf-edge/eden/pull/1019

PR #1025, opened 2024-09-13
Add virtualization workflow with no TPM

milan-zededa:
We want to test deploying container-in-VM to a device without TPM. In the past we had some bug merged because this was not tested. Also, it is better to use string attributes for workflow matrix, even for the boolean tpm option.

Then instead of workflow names:
```
eden-report-smoke.tests.txt-tpm-true-ext4
eden-report-smoke.tests.txt-tpm-false-zfs
```
We get somewhat more readable:
```
eden-report-smoke.tests.txt-tpm-ext4
eden-report-smoke.tests.txt-no-tpm-zfs
```
I'm open for one-word suggestions to replace "no-tpm". What EVE uses when TPM is not available? Golang crypto library? Should we call it "gocrypto"? I don't know.

### DIFF
--- a/.github/actions/run-eden-test/action.yml
+++ b/.github/actions/run-eden-test/action.yml
@@ -5,9 +5,9 @@ inputs:
   file_system:
     required: true
     type: string
-  tpm_enabled:
+  crypto_module:
     required: true
-    type: bool
+    type: string
   suite:
     required: true
     type: string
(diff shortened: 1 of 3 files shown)


milan-zededa:
Names of artifacts are not what I have hoped for, moving to draft..

PR #1031, opened 2024-09-24
Make eden great again

shjala:
This is really janky at the moment, but the idea is to get rid of the escrip and write tests using go testing infra.

1. Goal is to follow two principales, DRY and KISS.
2. Tests are grouped together (only smoke at the moment)
3. Tests broken into different categories based on what they test, like just EVE, or App or Network (for now just simple eve tests are ported)
4. Idea is setup once and use in multiple tests, for example setup and app once and share it between (no-destructive) tests.
5. This is WIP through and through. 
6. I'm going to gradually cherry pick relevant master commits to this branch.


shjala:
> You sure you want to do this in one PR? I mean I'm going to help you all the way through, but my experience with that says we should do it in smaller steps

hmm make sense, lets me rework it so we can have two eden in parallel.

uncleDecart:
@shjala since #1037 is merged makes sense to rebase this, but let's have a discussion first about evetestkit design

shjala:
> @shjala since #1037 is merged makes sense to rebase this, but let's have a discussion first about evetestkit design

sure 👍

shjala:
I'm going to add the setup and teardown to this and then lets have a final round of review, I don't want to add more tests to it to bloat it more that it is.

PR #1033, opened 2024-10-17
add eve_log_level parameter to GH test workflow

europaul:
It would be useful to be able to set the log level for EVE when running tests, as e.g. "debug" log level would provide more information for debugging failed tests.

It would be good to include this change in the next release of Eden, so that it can be used workflows started from EVE repository.

### DIFF
--- a/.github/actions/setup-environment/action.yml
+++ b/.github/actions/setup-environment/action.yml
@@ -10,6 +10,10 @@ inputs:
     type: bool
   eve_image:
     type: string
+  eve_log_level:
+    type: string
+    required: false
+    default: 'info'
   eve_artifact_name:
     type: string
   artifact_run_id:
@@ -90,6 +94,10 @@ runs:
         else
           echo "Skipping setting up eve image ${image}"
         fi
+      shell: bash
+      working-directory: "./eden"
+    - name: Set eve log level
+      run: ./eden config set default --key=eve.log-level --value=${{ inputs.eve_log_level }}
       shell: bash
       working-directory: "./eden"
     - name: Setup ext4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ on:  # yamllint disable-line rule:truthy
       eve_log_level:
         type: string
         required: false
-        default: 'info'
+        default: 'debug'
       eve_artifact_name:
         type: string
       artifact_run_id:
@@ -29,7 +29,7 @@ on:  # yamllint disable-line rule:truthy
       eve_log_level:
         type: string
         required: false
-        default: 'info'
+        default: 'debug'
       eve_artifact_name:
         type: string
       artifact_run_id:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,6 +10,10 @@ on:  # yamllint disable-line rule:truthy
         default: ''  # if not provided: When checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event. Otherwise, uses the default branch.
       eve_image:
         type: string
+      eve_log_level:
+        type: string
+        required: false
+        default: 'info'
       eve_artifact_name:
         type: string
       artifact_run_id:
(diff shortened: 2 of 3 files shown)


.github/actions/setup-environment/action.yml
```diff
@@ -92,6 +96,10 @@ runs:
         fi
       shell: bash
       working-directory: "./eden"
+    - name: Set eve log level
+      run: ./eden config set default --key=eve.log-level --value=${{ inputs.eve_log_level }}
```
giggsoff, line 100: As I remember there were two [options](https://github.com/lf-edge/eve/blob/master/docs/CONFIG-PROPERTIES.md) for tests: one to collect inside EVE-OS, another one to send to controller. Can you please validate, that your changes are enough?

europaul: thank you for the suggestion @giggsoff !

the comments in the [template](https://github.com/lf-edge/eden/blob/e974c0afc9b817350a2d15301b41b4954ca54dd2/pkg/defaults/templates.go#L157-L162) say it's the right field to change, since I want to have this log level to be captured in the artifacts that Eden workflows produce (in collect_logs mostly).

giggsoff: Thank you for validation!
Note, that If you say about the step [here](https://github.com/lf-edge/eden/blob/8545f10ca982fc9b621aab74616eb03f1a217e12/.github/actions/publish-logs/action.yml#L12), then they come from controller (Adam).

giggsoff: Ah, I see. It should work for extraction logs directly from device's store.
The store has its rotation, but it should be enough for tests for sure.

uncleDecart:
But how would you execute a workflow with debug log level in EVE? You want to introduce manual eden test run action?

europaul:
> But how would you execute a workflow with debug log level in EVE? You want to introduce manual eden test run action?

No, I would just set it for the existing eden test workflow for every PR. Like this https://github.com/lf-edge/eve/pull/4381

.github/workflows/test.yml
```diff
@@ -10,6 +10,10 @@ on:  # yamllint disable-line rule:truthy
         default: ''  # if not provided: When checking out the repository that triggered a workflow, this defaults to the reference or SHA for that event. Otherwise, uses the default branch.
       eve_image:
         type: string
+      eve_log_level:
+        type: string
+        required: false
+        default: 'info'
```
milan-zededa, line 16: Before using this from the eve repo, we should try to set 'debug' first here.
We need to check how large the output artifacts will be and whether they are not exceeding some github limits.

europaul: I couldn't find any official documentation on the topic, but there is a [rumor](https://github.com/actions/upload-artifact/issues/9) that the limit is 5 GB per file.

Our artifacts still stay under 50 MB https://github.com/lf-edge/eden/actions/runs/11437464614?pr=1033

europaul: before
after
Here are the before and after screenshots. I'd say the sizes stay comparable.

milan-zededa: Does it actually work?
When I look for example into https://github.com/lf-edge/eden/actions/runs/11437464614/artifacts/2081755978,
I do not see debug logs inside eve-info.tar.gz
`debug.default.loglevel` is configured to `info` in `persist-status/zedagent/ConfigItemValueMap/global.json`

europaul: Sorry, I just noticed that as well :facepalm: 
Will fix very soon

europaul: ok, now I think I got it right
image
the sizes have increased but still not dramatic

europaul: maybe we can try running it like this for a while and if we reach some limit we can still revert it

europaul: btw github copilot can provide some interesting insights that cannot be found in the github documentation
image
I'm just wondering if it's true or hallucinations :smile:

europaul:
folks, could you give this PR another look? I think I answered all the comments and fixed the issues.
let me know if it's good to merge and I'll remove the last two commits, which were here for demonstration purposes only.

milan-zededa (approved):
LGTM and I would suggest to keep [set log-level to debug for all tests](https://github.com/lf-edge/eden/pull/1033/commits/c1481ddb205a87c0653be714ed2df0b5b6579431) so that we have more logs also here in the eden repo. Just revert the last commit.

PR #1034, opened 2024-10-17
Support for go 1 23

europaul:
Keeping up with the good tradition to use go's internal APIs and update the dependencies by hand :smile: 

Doesn't really need to be merged until we move to go 1.23 in go.mod, but can be useful for people using 1.23 locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,9 @@
 # Test binary, built with `go test -c`
 *.test
 
+# make build-tests artifacts
+tests/**/eden.*.test*
+
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
--- a/tests/escript/go-internal/testscript/testing_1.23.go
+++ b/tests/escript/go-internal/testscript/testing_1.23.go
@@ -1,5 +1,5 @@
-//go:build go1.18 && !go1.23
-// +build go1.18,!go1.23
+//go:build go1.23
+// +build go1.23
 
 package testscript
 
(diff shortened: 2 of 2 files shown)


uncleDecart:
Should be irrelevant once #1037 is merged and escript is deprecated

europaul:
@uncleDecart you're getting rid of the whole escript directory, right?

uncleDecart:
> @uncleDecart you're getting rid of the whole escript directory, right?

Yes, we merge my PR, move tests to NeoEden and then remove all the old tests completely. It'll take time to do so, though

europaul:
@uncleDecart I just had a look at #1037 and you are keeping `escript/go-internal/testscript/testing_1.18.go` without basically any changes. Didn't you want to get rid of it?

europaul:
@uncleDecart I guess we can close this one as not relevant anymore, right?

europaul:
@uncleDecart  sorry, I just read the previous discussion and now I think that we should rather merge this PR, since NeoEden tests will take a looong time to complete

once they are merged however we call remove both `tests/escript/go-internal/testscript/testing_1.18.go` and `tests/escript/go-internal/testscript/testing_1.23.go`

uncleDecart:
We will have to bump version in all the GHA files + in go.mod itself if we want to merge this

europaul:
> We will have to bump version in all the GHA files + in go.mod itself if we want to merge this

no, we can keep using 1.22.
the problem is only if somebody has their environment setup with go version 1.23: then the device only has go's standard library in v1.23 and this breaks current Eden implementation.

PR #1035, opened 2024-10-21
GitHub Actions: switch to running Eden on GitHub runners

uncleDecart:
Recently GitHub bumped the runners to be as powerful as BuildJet ones, and they're all free. So let's try them out

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,6 +28,7 @@ on:  # yamllint disable-line rule:truthy
         type: string
 
 jobs:
+  
   smoke:
     continue-on-error: true
     strategy:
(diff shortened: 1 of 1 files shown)


uncleDecart:
Huh, takes 1hr to finish smoke [tests](https://github.com/lf-edge/eden/actions/runs/11442762694/job/31833845150?pr=1035), maybe we need to enable those large runners, will check it out

uncleDecart:
GitHub runners don't have HW support for Virtualization. That will slow down the whole thing... Not sure if switching makes sense now

uncleDecart:
Closing PR, since there's no reason to switch to GitHub Runners

Issue #828, opened 2022-12-05
[EPIC] Eden as library


uncleDecart:
# What:

We want to refactor Eden so that we can use that as a standalone library, i.e. 
Right now you can do this

```bash
$ eden config add default
$ eden setup
$ eden start
$ eden eve onboard
$ eden pod deploy …
```

But after you’ll be able to do this as well
```golang
func setup() {
  eveCfg, _ := edenhelper.LoadConfig(utils.DefaultConfigPath())
  edenhelper.SetupEden(*eveCfg)
  edenhelper.StartEden(*eveCfg)
  edenhelper.OnboardEve(eveCfg.Eve.CertsUUID)
  evehelper.PodDeploy(*eveCfg, …)
}
```
# Outcomes:
- Simplified, flexible and robust EVE testing
- Possibility for 3rd party developers to use eve as code (i.e. integration of EVE into FL framework as infrastructure provider). This will need some good documentation for developers to use it
- Significantly simplified EVE development. The try-fail-repeat cycle currently is taking tens of minutes, with heavy evolvement of the developers. With eden-lib, one can put together 10 lines of go, which would spin up eden, onboard the node, and run some load which tests what is currently under devlopment/debug.

# Roadmap
- #803 
- #827

uncleDecart:
@giggsoff @milan-zededa I propose use this space as discussion for further improvements and creating roadmap. I suggest we will close this EPIC once all of the issues on the roadmap will be merged. This list is to be increased. I'm copying Milan's comment on callback from #803 here:

> I realize that I come a bit late for the review, but I was wondering if it would be possible to structure the openevec library in a more OOP way. Currently, it has a flat structure - consisting of a bunch of exported functions, most of which takes the big structure EdenSetupArgs. It is not very clear what each function actually uses from the structure and how it is related to other functions.
> Without a good knowledge of EVE (and Eden as implemented before) I would have hard time writing a program based on openevec. First, I would not know where to start.
> 
> The config structure is good as it is but I would prefer to work with handles, i.e. pointers to structures with the context, structured in a tree hierarchy for example.
> There could be a top-level handle for the library :
> 
> ```golang
> type OpenEVE struct {
>     config *EdenSetupArgs
>    // etc.
> }
> 
> func Setup(Config *EdenSetupArgs) *OpenEVE
> ```
> (what "c" actually stands for in openevec?)
> 
[...]

PR #1037, opened 2024-10-23
Add first golang-only test 

uncleDecart:


uncleDecart:
The pitfall is that we still write config, couldn't figure this one out. Time to fight the my way through CI/CD  

image

uncleDecart:
This should close https://github.com/lf-edge/eden/issues/828

uncleDecart:
Need to change project in tests...

shjala:
This is a big one!

uncleDecart:
locally smoke tests passed, removed sec_test because it's now part of NeoEden. Let's see how pipeline will behave :crossed_fingers:

uncleDecart:
any zfs experts or any advice on how I can run things locally to debug this thing?

uncleDecart:
Okay, ZFS is not working, one smoke test failed on shutdown, I remember it was flaky. I'll look into zfs now, but it's getting closer

uncleDecart:
Seems promising, hitting [TOOMANYREQUESTS](https://github.com/lf-edge/eden/actions/runs/11613915058/job/32340866498?pr=1037#step:3:1842) on User app test

pkg/openevec/config.go
```diff
@@ -73,6 +82,7 @@ type AdamConfig struct {
 	CertsEVEIP  string `mapstructure:"eve-ip" cobraflag:"eve-ip"`
 	APIv1       bool   `mapstructure:"v1" cobrafalg:"force"`
 	Force       bool   `mapstructure:"force" cobraflag:"force"`
+	CA          string `mapstructure:"ca"`
```
europaul, line 85: it is used anywhere?

pkg/openevec/config.go
```diff
@@ -21,6 +22,12 @@ type EServerConfig struct {
 	Tag    string       `mapstructure:"tag" cobraflag:"eserver-tag"`
 	IP     string       `mapstructure:"ip"`
 	Images ImagesConfig `mapstructure:"images"`
+	EVEIP  string       `mapstructure:"eve-ip"`
```
europaul, line 25: is it used anywhere?

pkg/openevec/config.go
```diff
@@ -37,10 +44,12 @@ type EdenConfig struct {
 	EdenBin      string `mapstructure:"eden-bin"`
 	TestBin      string `mapstructure:"test-bin"`
 	TestScenario string `mapstructure:"test-scenario"`
+	Tests        string `mapstructure:"tests"`
```
europaul, line 47: is it used anywhere?

pkg/openevec/eden.go
```diff
@@ -366,29 +371,24 @@ func setupEdenScripts(cfg EdenSetupArgs) error {
 }
 
 func setupConfigDir(cfg EdenSetupArgs, eveConfigDir, softSerial, zedControlURL string, grubOptions []string) error {
-	if _, err := os.Stat(filepath.Join(cfg.Eden.CertsDir, "root-certificate.pem")); os.IsNotExist(err) {
```
europaul, line 369: why regenerate certificates on every setup? I'm not against that, just wondering why

cmd/edenConfig.go
```diff
@@ -10,7 +11,11 @@ import (
 )
 
 func newConfigCmd(configName, verbosity *string) *cobra.Command {
```
europaul, line 13: will all these functions be removed anyway once NeoEden tests are done?

cmd/edenSetup.go
```diff
@@ -23,9 +23,9 @@ func newSetupCmd(configName, verbosity *string) *cobra.Command {
 		Long:              `Setup harness.`,
 		PersistentPreRunE: preRunViperLoadFunction(cfg, configName, verbosity),
 		Run: func(cmd *cobra.Command, args []string) {
-			if err := openevec.ConfigCheck(*configName); err != nil {
-				log.Fatalf("Config check failed %s", err)
-			}
+			// if err := openevec.ConfigCheck(*configName); err != nil {
+			// 	log.Fatalf("Config check failed %s", err)
+			// }
```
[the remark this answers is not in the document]
europaul: then `ConfigCheck` also becomes obsolete

pkg/defaults/templates.go
```diff
@@ -428,3 +428,152 @@ const ParallelsDiskTemplate = `<?xml version='1.0' encoding='UTF-8'?>
         </Shot>
     </Snapshots>
 </Parallels_disk_image>`
+
+// DefaultEdenTemplate is configuration template for Eden
+const DefaultActivateShTemplate = `# This file must be used with "source bin/activate" *from bash*
```
europaul, line 433: do we need to remove `activate.sh.tmpl` and `activate.csh.tmpl`?

pkg/evetestkit/utils.go
```diff
@@ -728,3 +731,121 @@ func InitilizeTest(projectName string, options ...TestOption) (*EveNode, error)
...
+		Tests:     map[*device.Ctx]*testing.T{},
+		SdnClient: sdnClient,
+		WithSdn:   withSdn,
+	}
+	tstCtx.ProcBus = testcontext.InitBus(tstCtx)
+	return tstCtx, nil
+
+}
+
+func InitilizeTestFromConfig(projectName string, cfg *openevec.EdenSetupArgs, options ...TestOption) (*EveNode, error) {
```
europaul, line 780: can this be used to initialize tests in `lim_test.go` and `reboot_test.go`?

europaul (commented):
very nice big change! I still have a lot of questions, but they are not as relevant. I think my biggest suggestion would be to make the config a nice JSON and forget about all that reflect stuff.

But as long as the tests and the CI work, I'm happy :)

pkg/openevec/config.go
uncleDecart: Needed for backward compatibility with escript tests, some of them use it.

pkg/openevec/config.go
uncleDecart: Needed for backward compatibility with escript tests, some of them use it, plus that's how we understand which IP has been assigned to EVE :)

pkg/openevec/config.go
uncleDecart: Used in escript, won't need this once we remove it. Then we can just see if the field is used and remove it, where needed

cmd/edenConfig.go
uncleDecart: Not so sure about it,
talking about it on high-level: openevec is your Code As Infrastructure (think terraform) which allows you to describe your environment to deploy EVE, then you have your evetestkit which allows you to access certain states of EVE, Adam and then you have a test, which utilizes openevec and evetestkit to setup certain environment, run certain commands via controller and check EVE state. Openevec can also be used to setup a certain environment via CLI. 
So broadly speaking, I can see scenarios, where I'd like to keep configurations persistent and therefore want to keep it on disk. Talking about it practically, we can't remove it, because changers are dependent on default folder structure and files, in theory we could remove it, propogating configuration there, but that'd be another big refactoring patch, as of now I'd pursue this

pkg/defaults/templates.go
uncleDecart: oh yes, we should remove the files, they're not needed anymore, thx!

pkg/evetestkit/utils.go
uncleDecart: yup, this should be used in all "NeoEden" tests

uncleDecart: I'd submit a separate PR with that, this one is already big

pkg/utils/files.go
```diff
@@ -99,23 +99,28 @@ func TouchFile(src string) (err error) {
...
 }
 
-//ResolveAbsPath use eden.root parameter to resolve path
+// ResolveAbsPath use eden.root parameter to resolve path
 func ResolveAbsPath(curPath string) string {
+	return ResolveAbsPathWithRoot(viper.GetString("eden.root"), curPath)
+}
+
+// ResolveAbsPathWithRoot use rootPath parameter to resolve path
+func ResolveAbsPathWithRoot(rootPath, curPath string) string {
```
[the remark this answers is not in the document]
uncleDecart: There won't be trimming, which could lead to wrong path.

pkg/openevec/config.go
```diff
@@ -357,3 +367,93 @@ func ConfigCheck(configName string) error {
...
+
+func getValStrRepr(v reflect.Value) string {
+	if v.Kind() == reflect.String {
+		return fmt.Sprintf("'%v'", v.Interface())
+	} else {
+		return fmt.Sprintf("%v", v.Interface())
+	}
+}
+
+func WriteConfig(dst reflect.Value, root string, writer io.Writer, nestLevel int) {
```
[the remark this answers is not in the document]
uncleDecart: Backward compatibility with viper :) We need yaml, plus we need to trim some paths, also for backward compatibility, in theory once we remove rest of the dependencies, we can use default functions to marshal things. And granted, we can do it on a structure level, but it'll still be through tags and reflection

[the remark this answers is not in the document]
uncleDecart: I mean, of course you can specify a specific function,

```go
func (cfg *EdenSetupArgs) WithResolvedAbsPaths(root string) *EdenSetupArgs {
  cfg.Root.Eden = utils.ResolveWithAbsPath(root, cfg.Root.Eden)
  // ...
}m
```

In case of structure tags I specify how the struct should be marshalized and presented on disk and what is the operation I need to do when I'm reading from the disk.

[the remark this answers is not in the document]
uncleDecart: I guess you won't be surprised of my specific taste in those, since I do like C++ and template methods, however if there's a readability and maintainability benefit that I don't see I'm all for changing the approach :)

cmd/edenSetup.go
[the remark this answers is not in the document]
uncleDecart: :saluting_face:

pkg/openevec/config.go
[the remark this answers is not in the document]
europaul: but cannot we use https://pkg.go.dev/gopkg.in/yaml.v2 to marshal yaml?

cmd/edenConfig.go
europaul: I mean in more broader terms: can we remove the `eden` commands?

pkg/evetestkit/utils.go
```diff
@@ -728,3 +731,121 @@ func InitilizeTest(projectName string, options ...TestOption) (*EveNode, error)
...
+		return nil, err
+	}
+
+	// Registering our own project namespace with controller for easy cleanup
+	tc.InitProject(fmt.Sprintf("%s_%s", projectName, time.Now()))
+
+	// Create representation of EVE instances (based on the names
+	// or UUIDs that were passed in) in the context. This is the first place
+	// where we're using zcli-like API:
+	for _, node := range tc.GetNodeDescriptions() {
```
europaul, line 793: I'm a little confused by this function: why is it returning a slice, when Eden supports only one node?

cmd/edenConfig.go
europaul: but your overview is very nice, thank you :heart:

pkg/openevec/config.go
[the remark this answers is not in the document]
uncleDecart: we can, but there's still a thing with resolving paths, which can be solved either by providing function like WithResolvedAbsPaths and WithTrimmedAbsPaths where developed should specify fields, which should be resolved and then specify them to be unresolved, alternatively we can use struct tags and have generic function over fields which does that when marshalling and unmarshalling.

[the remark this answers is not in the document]
uncleDecart: I prefer struct tags, because you write them once and then you forget, whenever you need a resolved field, you just specify tag in struct, everything else is taken care of. When using functions it'll create boilerplate code with just setting specific variables and then unsetting them, granted, might be easier to read, but when you need to add a field, you need to add that field in two functions. I personally think this is not clean. I do acknowledge that this is matter of taste, so I'm fine with changing it if there are problems with code understanding and/or readability. Is there something particularly nasty about this function in your opinion? Apart from the fact that it does something very similar to standard library?

pkg/openevec/config.go
```diff
@@ -179,6 +188,7 @@ type EdenSetupArgs struct {
 
 	ConfigFile string
 	ConfigName string
+	EdenDir    string
```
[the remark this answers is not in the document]
uncleDecart: I wouldn't want that. DefaultEdenDir() uses some default environment variables. I would prefer to have it stored in a config which is passed as sort of context through tests, with that when I'm setting up test I get WYSIWYG with config: here is default config, here's some folders, yada yada yada. I might write something like this:

```go
cfg := openevec.DefaultConfig()
   .WithHypervisor(openevec::HyperVisorQemu)
   .WithApplicationInstances(openevec::DefaultApplication)
   .WithHomeDir("/my/awesome/dir")
```

and I'd like to guarantee that when I run test I will have this environment. With DefaultEdenDir() I take something from somewhere under assumption of some environment variables. In my opinion such things might be harder to debug or keep in mind (imagine, somewhere in the code we can use default, or the value you specified in config) 

image

cmd/edenConfig.go
uncleDecart: hmmm, that's an interesting take... Do we want Eden commands at all? I'd say yes. Imagine this: you run tests and then it gets stuck or fails, then in order to investigate you just set it up to a certain condition and then you can use eden commands to play around, ssh into thing, change certificates, etc. etc. so you don't waste time writing one openevec command and then  rerunning the whole thing, rather do it interactively

pkg/evetestkit/utils.go
uncleDecart: well, there was support for multinodes, the whole point of contexts was about multi-node support, if I remember it correctly

uncleDecart: but tc is actually next thing to refactor IMO, it's what's been used under the hood in escript, but now with simplified context of running (just one golang program) and the whole toolkit of golang, definitions, guidlines, books, etc etc. 

However, in this particular case I don't think it's a problem. We define interface, which we ideally want to have (i.e. that nice little black box which creates N edge nodes, sets up everything, makes really nice coffee, etc.) then we deal with implementation based on our sense of urgency of the features.

Granted, it's hard to implement ideal interface, that's why it's iterative process, but it's usually harder to change interface than an implementation which follows the same interface (or part of it)

pkg/openevec/config.go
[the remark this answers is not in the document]
europaul: makes sense. but you'll still probably store that config as a file in `EdenDir` so it's a little chicken-egg'y, but that's more of a side effect :)

pkg/openevec/config.go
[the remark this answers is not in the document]
europaul: backward compatibility - touche, but as soon as we don't have that anymore I'd prefer to resort to some standard marshalizer. the reason is simple - simplicity :smile: if we just used a standard serializer we wouldn't be having this discussion. also it makes serializing less error prone, since I don't need to have special knowledge about this function to understand how the serialization happens so there is less chance for errors.

pkg/evetestkit/utils.go
europaul: but even our openEVEC supports only one EVE config...
would be good to clean it up in the future - either full multi-node support or just single node

pkg/openevec/config.go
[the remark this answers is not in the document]
uncleDecart: So you'd rather have something like this? 

```go
func ResolveAbsPath(dst reflect.Value, root string) {
	if dst.Kind() == reflect.Ptr {
		dst = dst.Elem()
	}

	for i := 0; i < dst.NumField(); i++ {
		if structTag := dst.Type().Field(i).Tag.Get("mapstructure"); structTag != "" {
			f := dst.Field(i)
			fieldType := dst.Type().Field(i)

			switch f.Kind() {
			case reflect.Struct:
				// Pass the addressable value of the struct if it can be set, else create a pointer and pass
				if f.CanAddr() {
					ResolveAbsPath(f.Addr(), root)
				} else {
					ResolveAbsPath(f, root)
				}
			case reflect.String:
				if _, ok := fieldType.Tag.Lookup("resolvepath"); ok {
					f.SetValue(strings.TrimPrefix(val, root+"/"))
				}
 			}
		}
	}
}

func (cfg* EdenSetupArgs) WriteToFile(path string) {
   cfgCopy := copy(cfg)

   ResolveAbsPath(&cfgCopy, cfg.Eden.Root)

   jsonData, err := json.Marshal(cfgCopy)
  if err != nil {
    fmt.Println("Error marshalling to JSON:", err)
    return
  }
  // .. 
}

```

pkg/openevec/config.go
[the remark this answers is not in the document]
uncleDecart: Well, EdenDir will be set for the config, chicken egg solves here in a manner config comes first :)

pkg/openevec/config.go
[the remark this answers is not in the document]
europaul: looks good - we can also put a comment that `ResolveAbsPath` is there for backward compatibility and can be removed once that's not needed anymore.

also I think if you pass param by value you can save the copy statement:
```golang
func (cfg EdenSetupArgs) WriteToFile(path string) {
```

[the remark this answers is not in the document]
uncleDecart: okay, makes sense

uncleDecart:
I'm creating a separate commit for review requests, I think it makes sense to squash all the commits after merging. Plus I added the doc explaining NeoEden high level

docs/NEO-EDEN.md
```diff
@@ -0,0 +1,199 @@
+# NeoEden: framework that allows you to write EVE tests in Golang
```
deitch, line 1: This is a really big improvement. I am going to put some comments and suggestions inline.

The main one, though, is to separate the friendly and fun, "what is neoeden and why do we need it and what was the history" (first ~92 lines), which will become more of a state-in-time curiosity once everything else is converted, from a "how to test with eden" doc (line 93 to end), which will continue to be up to date.

docs/NEO-EDEN.md
```diff
@@ -0,0 +1,199 @@
...
+
+**But how do you debug that test?** Because you're running an interpreter, which runs bash commands via os.exec plus you execute golang program, which is written and compiled inside the repository. Debug printing would work, but I find debuggers much more useful and efficient most of the times. It is a matter of taste, but better to debugger at your disposal, than not to have it in this case. Even worse, how do you debug escript problem? You need to know it's internals, there's no Stackoverflow or Google by your side, nor there are books about it.
+
+**Bumping golang is actually manual action**: last but not least, we have to maintain custom test files which a basically copy-paste with one added function, bumping golang turns into some weird dances in that case.
+
+I believe that those reasons are good enough to switch for golang-only approach to write tests, which is called NeoEden (kudos to @shjala for the name)
+
+Also worst case what could happen with DSL is something like the Inner JSON effect [link](https://thedailywtf.com/articles/the-inner-json-effect)
+
+## Enter NeoEden. How tests look now?
```
deitch, line 93: This would be the beginning of a standalone document that describes how to test with eden.

docs/NEO-EDEN.md
```diff
@@ -0,0 +1,199 @@
...
+
+They look like [this](../tests/sec/sec_test.go) code listing is in the end of this section. Conceptually, Eden contains of two entities: an entity which allows you setup some kind of environment and deploy EVE there. Think Infrastructure as a Code, like Teraform, which allows me using golang code to deploy, this is openevec package; second entity is evetestkit, which serves as an abstraction over missing features of adam as controller such as keeping track of state and provide you with toolkit to test things, namely send commands and access EVE or its AppInstances to see if the certain file exists or certain request gives you certain response. Those two things let you create golang-only test like one below. You use openevec to setup environment and evetestkit to help you describe what is passing criterion for particular test. The common denominator between openevec and evetestkit is EdenSetupArgs, openevec is just a wrapper over that configuration, which defines your state. The goal is to be able to define your state, write custom functions to state, like this:
```
deitch, line 95: Can you start with an overview (unless it is elsewhere, e.g. in another doc, and I missed it)? Something like:

In order to test eve with eden, there are a few steps you will need to take:

1. Start backends, controller, etc.
2. Read config
3. etc. etc.

Get the major ones.

Then describe what the golang functions are for those. 

Then go to an example. Or, you can put the example at the beginning, but explain that you will describe it step by step afterwards. In the example, it should be described step-by-step, maybe inline comments before each action describing what it does. Make it easy for someone reading it to understand.

deitch: In terms of the content of this paragraph itself:

> Conceptually, Eden contains of two entities: an entity which allows you setup some kind of environment and deploy EVE there. Think Infrastructure as a Code, like Teraform, which allows me using golang code to deploy, this is openevec package; second entity is evetestkit, which serves as an abstraction over missing features of adam as controller such as keeping track of state and provide you with toolkit to test things, namely send commands and access EVE or its AppInstances to see if the certain file exists or certain request gives you certain response.

I wouldn't want to reference Terraform explicitly. Besides trademark issues, people will think this is much more than it is and might confuse them.

Maybe something like:

```
The eden golang test SDK for eve consists of two categories of functions:

1. `openevec` (link to the godoc for it): infrastructure manager. `openevec` enables you to deploy a controller, a backend for that controller, deploy an edge node, etc. (what else does it do?). It does not change anything on an individual edge node.
2. `evetestkit` (link to the godoc for it): provides fine-grained control over specific functions on individual edge nodes, like deploying applications or network instances, or deleting them. 

There is more you have, but it provides an entry structure that will make understanding the documented example easier.

File: docs/NEO-EDEN.md [hunk 16]
```diff
@@ -0,0 +1,199 @@
...
+  .WithAppInstances(appInstance)
+  .WithPatchEnvelopes(pe)
+
+evec := openvec.CreateOpnEVEC(cfg)
+
+evec.SetupEden(/* ... */)
+evec.StartEden(/* ... */)
+evec.OnboardEve(/* ... */)
+
+node, err := tk.InitializeTestFromConfig(cfg)
```
deitch, line 114: What is `tk` here? Is this `node` part of `openevec` or `evetestkit`?

deitch (changes requested):
I commented just on the docs. Let me know if this is helpful.

docs/NEO-EDEN.md
uncleDecart: evetestkit

docs/NEO-EDEN.md
uncleDecart: well, openevec allows you to deploy applications and to be fair, access node as well, I'd say evetestkit is a framework to easily describe what is the expected result for the test

docs/NEO-EDEN.md
deitch: Got it. Should be clearer.

docs/NEO-EDEN.md
deitch: So my docs were off. General idea, though, would be really helpful.

uncleDecart:
@deitch second pass on docs :)

docs/writing-tests.md
```diff
@@ -0,0 +1,99 @@
...
+
+And those functions are used in context of standard golang test library. We do a setup in TestMain and then write test functions which interact with the environment we created. Setting up environment takes couple of minutes, so it makes sense to do it once and run tests within that environment. 
+
+Source for the example below can be found [here](../tests/sec/sec_test.go)
+
+In order to test a feature in Eden you need to
+
+## 1. Create a configuration file and describe your environment
+
+The glue between `openevec` and `evetestkit` is a configuration structrure called `EdenSetupArgs`. It contains all the necessary information to setup all the components of the test: controller, EVE, etc. You can fill in the structure manually, however `openevec` provides you with a convinient function to create default configuration structure providing project root path:
```
deitch, line 16: - "structure" (typo)
- "convenient" (typo)

pkg/evetestkit/utils.go
giggsoff: @uncleDecart is right, the context is about an instance of EVE-OS which may be connected to the same Adam. Another context will use its own set of certificates to not conflict on Adam side. And we can switch between contexts from CLI. Nice to keep the ability.

docs/writing-tests.md
```diff
@@ -0,0 +1,99 @@
...
+	cfg := openevec.GetDefaultConfig(twoLevelsUp)
+
+	if err = openevec.ConfigAdd(cfg, cfg.ConfigName, "", false); err != nil {
+		log.Fatal(err)
+	}
+  // ... 
+}
+```
+
+Due to backward compatibility with escript as of time of writing, the project root path should be Eden repository folder. So if you add tests in the `eden/tests/my-awesome-test` you need to go two levels up.
```
deitch, line 36: I assume this also is planned to be removed, and will be once all of the escript tests are gone? If so, worth saying explicitly.

docs/writing-tests.md
```diff
@@ -0,0 +1,99 @@
...
+
+`evetestkit` provides an abstraction over EveNode, which is used to describe expected state of the system. Each EveNode is running within a project You can create a global object within one test file to use it across multiple tests.
+
+```go
+const projectName = "security-test"
+var eveNode *evetestkit.EveNode
+
+func TestMain(m *testing.M) {
+  // ...
+	node, err := evetestkit.InitilizeTestFromConfig(projectName, cfg, evetestkit.WithControllerVerbosity("debug"))
```
deitch, line 67: I assume this is a typo and is meant to be `InitializeTestFromConfig`?

docs/writing-tests.md
```diff
@@ -0,0 +1,99 @@
...
+
+	eveNode = node
+	res := m.Run()
+	os.Exit(res)  
+}
+```
+
+## 4. Write your test
+
+Below is an example of test, which check if AppArmor is enables (specific file on EVE exists). It uses `EveReadFile` function from `evetestkit`
```
deitch, line 80: - "is enabled" (typo)

docs/writing-tests.md
```diff
@@ -0,0 +1,99 @@
...
+
+```go
+const appArmorStatus = "/sys/module/apparmor/parameters/enabled"
```
deitch, line 83: It would be great if this (and many other files) were parametrized, so they could be a property of some library. But that is beyond the scope of here.

.github/actions/setup-environment/action.yml
```diff
@@ -105,3 +106,10 @@ runs:
         ./eden setup -v debug --grub-options='set_global dom0_extra_args "$dom0_extra_args eve_install_zfs_with_raid_level "'
       shell: bash
       working-directory: "./eden"
+
+    - name: Start and Onboard
```
giggsoff, line 110: Looks like you disable setup in workflows and move its logic here. Do you have any reason for that?

giggsoff: Will we keep our collect-info and other debugging logic in case of failure on onboarding stage, which is quite predictable?

deitch (changes requested):
Much better. I can read the docs without too much prior knowledge and get a feel how to do things.

I made some suggestions related to typos and clarifying minor things.

Additional points:

* what about teardown? Don't I need to use `evec` for that? Shouldn't that be at the end of `TestMain`?
* People using this might (or might not) be used to `TestMain`, so it is worth putting in the `m.Run()` in the example. In that light, maybe link to the [TestMain official docs](https://pkg.go.dev/testing#hdr-Main)?
* Is there a better way to pass the node around than is a package-level variable? If not, ok, but worth asking. 
* Are there concerns about tests run in parallel? I would think they conflict with a single controller, so it might be worth stating explicitly not to invoke any `.Parallel()`
* I still am a little confused at the border between `openevec` and `evetestkit`. It is related to the context-passing question. I see very clearly how I use `openevec` to initialize things. But then I use `evetestkit` to get a node, and use that node to do things on it. Would "get a node" be part of `openevec`, and then `evetestkit` is just a set of convenient library functions? I am not sure I get where the border is.

.github/actions/setup-environment/action.yml
uncleDecart: Hey @giggsoff , nice seeing you here :) So I did it purely from backward compatibility. For some reason escript tests didn't want to onboard through test. 

Yes, we keep all collect-info logic there, no changes in that regard

uncleDecart:
> what about teardown? Don't I need to use evec for that? Shouldn't that be at the end of TestMain?

We can execute openevec.EdenClean(), ideally we should discuss how we should proceed with it. Because technically make clean does more than EdenClean(), it removes files as well, we need to fix it separately, this vicious cycle shouldn't be there (i.e. you should have teardown in tests). Hope that clarifies a bit :D 


>    People using this might (or might not) be used to TestMain, so it is worth putting in the m.Run() in the example. In that light, maybe link to the [TestMain official docs](https://pkg.go.dev/testing#hdr-Main)?

Done

> Is there a better way to pass the node around than is a package-level variable? If not, ok, but worth asking.

It's a good question: I would actually gather people to chat about evetestkit and how better to structure things there.

> Are there concerns about tests run in parallel? I would think they conflict with a single controller, so it might be worth stating explicitly not to invoke any .Parallel()

Nice catch! Done

> I still am a little confused at the border between openevec and evetestkit. It is related to the context-passing question. I see very clearly how I use openevec to initialize things. But then I use evetestkit to get a node, and use that node to do things on it. Would "get a node" be part of openevec, and then evetestkit is just a set of convenient library functions? I am not sure I get where the border is.

The problem is we haven't defined the border and now would be a really good time to discuss where this border should be.

On an all I don't now if this PR is the best place to do so, but I want to talk about design of golang test framework and two entities constituting Eden and what separation would be the best. We surely could (and I think should) start here to get some pointers on that discussion.

My take: openevec is about "basic" operations, which is: create EVE image, start controller, start EVE, enable networking with specific configuration, deploy specific App to EVE, ssh to EVE. Sort of create a setup and provide tools to access to components of that configuration based on configuration itself (I want to ssh to EVE, but actually if SDN is enabled I need to forward my request through it I want this logic to be hidden behind openevec, where I just say ssh to EVE; same applies to access controller, could be adam, could be commercial controller, could be anything else, I just want to access controller)

Then there's evetestkit which utilizes in some way openevec and provide functions of convenience, like ReadFile (which is actually Ssh to EVE and run cat command) or check that this App on Controller is in UP state (which is actually access controller, how to do that is hidden behind openevec)

uncleDecart:
@deitch third take on docs answered your questions above ^^^

deitch:
> It's a good question: I would actually gather people to chat about evetestkit and how better to structure things there.

It really isn't part of the question related to this, as it is to passing variables around between tests. When you have a `TestMain`, and want to pass things around to individual tests when you do `m.Run()`, how do you do it? If there is no clean answer right now, let it go for now. It can be addressed later, and isn't fundamental to what you have done here.

> The problem is we haven't defined the border and now would be a really good time to discuss where this border should be.
>I want to talk about design of golang test framework and two entities constituting Eden and what separation would be the best. We surely could (and I think should) start here to get some pointers on that discussion.

Yup, I think you started to do so in a good way:

> My take: openevec is about "basic" operations, which is: create EVE image, start controller, start EVE, enable networking with specific configuration, deploy specific App to EVE, ssh to EVE. Sort of create a setup and provide tools to access to components of that configuration based on configuration itself (I want to ssh to EVE, but actually if SDN is enabled I need to forward my request through it I want this logic to be hidden behind openevec, where I just say ssh to EVE; same applies to access controller, could be adam, could be commercial controller, could be anything else, I just want to access controller)
>
> Then there's evetestkit which utilizes in some way openevec and provide functions of convenience, like ReadFile (which is actually Ssh to EVE and run cat command) or check that this App on Controller is in UP state (which is actually access controller, how to do that is hidden behind openevec)

So is it correct to say that `openevec` is all things that I do in tests, whether infrastructure type stuff (set up eden, start the controller, start a virtual device, onboard a device, etc. etc.), and `evetestkit` is just ease-of-use utilities? I can do everything without evetestkit, if I want, it is just a bit of a pain, I have to do lots of lower-level commands repetitively. With `evetestkit`, I get a bunch of convenience functions. `evetestkit` doesn't talk to the filesystem, doesn't know about sockets or infrastructure or anything directly. It just knows how to use the lower-level functions/structs/interfaces provided by openevec, or structs/interfaces generated by it (e.g. a node) to do my stuff.

If that is correct - and I think that makes sense, although the naming "evetestkit" is a bit misleading, but not so far that it is worth changing (and I don't have a better name, so I have no right to complain), then I think it is a pretty solid line.

If that is true, though, then evetestkit should not depend on anything local or direct access to anything. The only parameters to any of its functions should be one of:

* something returned by openevec
* configuration (like the name of the app to deploy, or the file to check, etc.)
* something itself returned by evetestkit (to build higher level utilities)

If that is true, then it is a clean, easy to understand division of labour.

uncleDecart:
>>    It's a good question: I would actually gather people to chat about evetestkit and how better to structure things there.

> It really isn't part of the question related to this, as it is to passing variables around between tests. When you have a TestMain, and want to pass things around to individual tests when you do m.Run(), how do you do it? If there is no clean answer right now, let it go for now. It can be addressed later, and isn't fundamental to what you have done here.

Seems like it's a common [practice](https://stackoverflow.com/a/42395998)

Regarding the separation: we need to move some of the functions from there for it to be able to utilize openevec only. I suggest we also get input from @shjala and do it in a separate PR, I don't want to bloat this one, let's iterate, it's easier

deitch:
> Seems like it's a common [practice](https://stackoverflow.com/a/42395998)

Doesn't make it a good practice. 😆
Either way, don't worry about it now. It works.

> Regarding the separation: we need to move some of the functions from there for it to be able to utilize openevec only. I suggest we also get input from @shjala and do it in a separate PR, I don't want to bloat this one, let's iterate, it's easier

OK, as long as we get it done. The whole point of your excellent work here is to make it easier to use and understand.

pkg/openevec/eden.go
uncleDecart: Reverted that :)

pkg/evetestkit/utils.go
```diff
@@ -728,3 +731,121 @@ func InitilizeTest(projectName string, options ...TestOption) (*EveNode, error)
...
+		SdnClient: sdnClient,
+		WithSdn:   withSdn,
+	}
+	tstCtx.ProcBus = testcontext.InitBus(tstCtx)
+	return tstCtx, nil
+
+}
+
+func InitilizeTestFromConfig(projectName string, cfg *openevec.EdenSetupArgs, options ...TestOption) (*EveNode, error) {
+	var edgenode *device.Ctx
```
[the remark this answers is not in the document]
uncleDecart: yup, done

pkg/utils/params.go
```diff
@@ -1,12 +1,29 @@
...
 	"regexp"
 	"strings"
 	"time"
 )
 
+// GetControllerMode parse url with controller
+func GetControllerMode(controllerMode, modePattern string) (modeType, modeURL string, err error) {
+	params := GetParams(controllerMode, modePattern)
+	if len(params) == 0 {
+		return "", "", fmt.Errorf("cannot parse mode (not [file|proto|adam|zedcloud]://<URL>): %s", controllerMode)
```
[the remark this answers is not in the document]
uncleDecart: makes sense, done

pkg/openevec/config.go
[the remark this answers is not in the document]
uncleDecart: Let's do this change last, because in order to marshal to yaml we need to provide structure tags to it, I'd very much like to have only one tag to marshal things. I'd like to have it in a separate PR, describing possible backward compatibility issues and all, I'll create an issue mentioning this comment rn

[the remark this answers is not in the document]
europaul: sure, makes sense

uncleDecart:
I think I addressed all the comments, @deitch if you don't have any comments let's merge it after releasing 0.9.13 (we need hotfix for Eden tests)

deitch:
We still have work to do to make it cleaner separation for evetestkit, but if I understand you correctly, that is beyond the scope of this PR, so good.

One final request:

```go
func InitilizeTest()
func InitilizeTestFromConfig()
```

Can you please fix these typos?

```go
func InitializeTest()
func InitializeTestFromConfig()
```

It confuses people, hard to search, and doesn't make us look as good.

Other than that, good to go.

uncleDecart:
Shoot, missed the typo in the original function I guess, thanks Avi!

uncleDecart:
@deitch fixed the typo

pkg/evetestkit/utils.go
```diff
@@ -663,7 +666,7 @@ func GetRandomAppName(prefix string) string {
 func InitilizeTest(projectName string, options ...TestOption) (*EveNode, error) {
```
deitch, line 666: Can you fix this one too @uncleDecart ?

uncleDecart: Gosh, weird that you couldn't see my local changes that I haven't pushed :D Now it's fixed

deitch: Somehow, my AI vision didn't pick up on that. 🤷‍♂️

deitch (approved):

deitch:
yetus is unhappy, but other than that LGTM

uncleDecart:
I see now only complaint to bump GHA from Yetus now

uncleDecart:
Merging this, yetus complaints are about GHA bumps. Escript is deprecated. May the force be with us

uncleDecart:
Not at all @milan-zededa, we need to port all those escipt tests to new fancy golang framework, want to help? :D

milan-zededa:
> Not at all @milan-zededa, we need to port all those escipt tests to new fancy golang framework, want to help? :D

That's a very tempting offer, I'll think about it ;)

PR #1038, opened 2024-10-24
Set App virtualization mode via PodConfig

shjala:
This is needed, for example for loading FML based windows VMs.

### DIFF
--- a/pkg/openevec/config.go
+++ b/pkg/openevec/config.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/lf-edge/eden/pkg/defaults"
 	"github.com/lf-edge/eden/pkg/utils"
+	"github.com/lf-edge/eve-api/go/config"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
@@ -210,6 +211,7 @@ type PodConfig struct {
 	OpenStackMetadata bool
 	DatastoreOverride string
 	ACLOnlyHost       bool
+	VirtMode          config.VmMode
 }
 
 func Merge(dst, src reflect.Value, flags *pflag.FlagSet) {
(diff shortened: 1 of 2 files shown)


uncleDecart:
Do we need any changes on Adam side?

pkg/openevec/config.go
```diff
@@ -210,6 +211,7 @@ type PodConfig struct {
 	OpenStackMetadata bool
 	DatastoreOverride string
 	ACLOnlyHost       bool
+	VirtMode          config.VmMode
```
giggsoff, line 214: Would be great to add related `virt-mode` flag into [podDeployCmd](https://github.com/lf-edge/eden/blob/f81d9918ecdb5532c14f59ae4f4db7603cc74b0a/cmd/edenPod.go#L126) to support it in CLI.

shjala: sure!

shjala:
> Do we need any changes on Adam side?

I don't think so, the `expect` already sends this, but before this change it was the default virt mode. I have a test Windows VM test that works with just these changes (I mean EVE gets the virt-mode, which in my case is FML).

pkg/openevec/config.go
shjala: Added, is the default value is correct? or should be something else?

giggsoff: I am not sure, please [check](https://github.com/lf-edge/eden/blob/f81d9918ecdb5532c14f59ae4f4db7603cc74b0a/pkg/expect/expectation.go#L142-L149).

shjala: I set the default according to what you provided.

giggsoff: Looks like some tests failed with `level=fatal msg="Unknown virt-mode: "` (virtualization test). Can you please try to deploy something locally?

shjala: shoot, sorry.

PR #1045, opened 2024-12-02
gha: add neoeden job to test.yml workflow

uncleDecart:
Instead of adding additional workflow to eden, we run neoeden alongside with older version intent is that end user (EVE CI/CD pipeline) will not require any changes, it will still call test.yml there will be an additional test suite. End goal is to gradually move all the test suites to golang-based tests, so at some point smoke tests will be running in neoeden, or networking tests, etc. etc.

### DIFF
--- a/.github/actions/run-neoeden-test/action.yml
+++ b/.github/actions/run-neoeden-test/action.yml
@@ -28,6 +28,8 @@ inputs:
     type: string
     required: false
     default: ''
+  test_suite:
+    type: string
   aziot_id_scope:
     description: 'Azure IoT ID scope'
     required: false
@@ -60,14 +62,10 @@ runs:
         eve_artifact_name: ${{ inputs.eve_artifact_name }}
         artifact_run_id: ${{ inputs.artifact_run_id }}
         require_virtualization: ${{ inputs.require_virtualization }}
-    - name: Run kubevirt tests
-      run: go test
+    - name: Run neoeden test suite
+      run: go test -timeout 30m
       shell: bash
-      working-directory: "./eden/tests/kubevirt"
-    - name: Run security tests
-      run: go test
-      shell: bash
-      working-directory: "./eden/tests/sec"
+      working-directory: ${{ inputs.test_suite }}
       env:
         AZIOT_ID_SCOPE: ${{ inputs.aziot_id_scope }}
         AZIOT_CONNECTION_STRING: ${{ inputs.aziot_connection_string }}
@@ -87,4 +85,3 @@ runs:
         docker system prune -f -a >/dev/null
       shell: bash
       working-directory: "./eden"
-
--- a/.github/actions/run-neoeden-test/action.yml
+++ b/.github/actions/run-neoeden-test/action.yml
@@ -1,0 +1,90 @@
+name: 'Run Neo-Eden test workflow'
+description: 'Setup Eden, run go test and publish logs'
+
+inputs:
+  file_system:
+    required: true
+    type: string
+  tpm_enabled:
+    required: true
+    type: bool
+  eve_image:
+    type: string
+  eve_log_level:
+    type: string
+    required: false
+    default: 'info'
+  eve_artifact_name:
+    type: string
+  artifact_run_id:
+    type: string
+  require_virtualization:
+    type: bool
+  docker_account:  # if not provided: use anonymous docker user
+    type: string
+    required: false
+    default: ''
+  docker_token:
+    type: string
+    required: false
+    default: ''
+  aziot_id_scope:
+    description: 'Azure IoT ID scope'
+    required: false
+  aziot_connection_string:
+    description: 'Azure IoT connection string'
+    required: false
+
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Collect Workflow Telemetry
+      uses: catchpoint/workflow-telemetry-action@v2
+      with:
+        proc_trace_sys_enable: true
+        comment_on_pr: false
+    - name: Login to Docker Hub
+      if: inputs.docker_account != ''
+      uses: docker/login-action@v3
+      with:
+        username: ${{ inputs.docker_account }}
+        password: ${{ inputs.docker_token }}
+    - name: Setup Environment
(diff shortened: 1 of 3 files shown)


europaul:
@uncleDecart but do we really need this workflow? I thought that golang tests are just compiled and run with the rest of Eden tests in the regular `test` workflow

.github/actions/run-neoeden-test/actions.yml
```diff
@@ -0,0 +1,86 @@
...
+        tpm_enabled: ${{ inputs.tpm_enabled }}
+        eve_image: ${{ inputs.eve_image }}
+        eve_log_level: ${{ inputs.eve_log_level }}
+        eve_artifact_name: ${{ inputs.eve_artifact_name }}
+        artifact_run_id: ${{ inputs.artifact_run_id }}
+        require_virtualization: ${{ inputs.require_virtualization }}
+    - name: Run security tests
+      run: cd tests/sec && go test
+      shell: bash
+      working-directory: "./eden"
```
europaul, line 66:
```suggestion
      run: go test
      shell: bash
      working-directory: "./eden/tests/sec"
```
will it work like this? :thinking: then you don't need to change the directory back

.github/actions/run-neoeden-test/action.yml
```diff
@@ -0,0 +1,86 @@
...
+      with:
+        file_system: ${{ inputs.file_system }}
+        tpm_enabled: ${{ inputs.tpm_enabled }}
+        eve_image: ${{ inputs.eve_image }}
+        eve_log_level: ${{ inputs.eve_log_level }}
+        eve_artifact_name: ${{ inputs.eve_artifact_name }}
+        artifact_run_id: ${{ inputs.artifact_run_id }}
+        require_virtualization: ${{ inputs.require_virtualization }}
+    - name: Run security tests
+      run: cd tests/sec && go test
```
europaul, line 64: kudos to @shjala for the following comment:
why don't we put all commands calling the tests in this workflow in a bash script and call that script from here

Advantages:
- you don't need to change GHA every time you add a new test - you add it to that workflow bash script instead
- your newly added tests run immediately on the PR you added them in, because there is no change to the GHA itself

Disadvantages:
- some security concerns about the latter advantage and how people might change the bash script on their PR

.github/actions/run-neoeden-test/actions.yml
uncleDecart: Updated it :)

.github/workflows/test.yml
```diff
@@ -64,8 +64,16 @@ jobs:
...
+      - name: Check for DockerHub credentials
+        id: check-dockerhub-creds
+        run: |
+          if [ -n "${{ secrets.DOCKERHUB_PULL_USER }}" ] && [ -n "${{ secrets.DOCKERHUB_PULL_TOKEN }}" ]; then
+            echo "creds_exist=true" >> $GITHUB_OUTPUT
+          else
+            echo "creds_exist=false" >> $GITHUB_OUTPUT
+          fi
+      - name: Login to DockerHub
+        if: steps.check-dockerhub-creds.outputs.creds_exist == 'true'
```
europaul, line 80: chatGPT tells me you could do something like
```
- name: Login to DockerHub
  if: ${{ secrets.DOCKERHUB_PULL_USER != '' && secrets.DOCKERHUB_PULL_TOKEN != '' }}
  uses: docker/login-action@v3
  with:
    username: ${{ secrets.DOCKERHUB_PULL_USER }}
    password: ${{ secrets.DOCKERHUB_PULL_TOKEN }}
```

Idk if that really works, but seems leaner than grepping through the output of the previous step

.github/workflows/test.yml
```diff
@@ -382,3 +387,49 @@ jobs:
...
+          repository: "lf-edge/eden"
+          ref: ${{ inputs.eden_version }}
+          path: "./eden"
+      - name: Run Neoeden tests
+        uses: ./eden/.github/actions/run-neoeden-test
+        with:
+          test_suite: "./eden/tests/kubevirt"
+          file_system: ${{ matrix.file_system }}
+          tpm_enabled: ${{ matrix.tpm }}
+          eve_image: ${{ inputs.eve_image }}
```
europaul, line 428: shouldn't this be `inputs.eve_kubevirt_image`?

.github/workflows/test.yml
uncleDecart: You'd expect this to work, but it doesn't unfortunately :c

.github/actions/run-neoeden-test/action.yml
```diff
@@ -0,0 +1,87 @@
...
+      with:
+        file_system: ${{ inputs.file_system }}
+        tpm_enabled: ${{ inputs.tpm_enabled }}
+        eve_image: ${{ inputs.eve_image }}
+        eve_log_level: ${{ inputs.eve_log_level }}
+        eve_artifact_name: ${{ inputs.eve_artifact_name }}
+        artifact_run_id: ${{ inputs.artifact_run_id }}
+        require_virtualization: ${{ inputs.require_virtualization }}
+    - name: Run neoeden test suite
+      run: go test
```
andrewd-zededa, line 66: @uncleDecart this may need a timeout configured for go test otherwise the default go test will panic if it goes over 10min

> % go help testflag
> ...
>	-timeout d
> 	    If a test binary runs longer than duration d, panic.
> 	    If d is 0, the timeout is disabled.
> 	    The default is 10 minutes (10m).

andrewd-zededa: I've seen the kubevirt test go over this a few minutes and the panic stack dump to the console is surprisingly long.

andrewd-zededa (commented):
Interested in your thoughts on the timeout I mentioned.

andrewd-zededa:
Looks like the Networking test suite failed due to "TOOMANYREQUESTS: You have reached your unauthenticated pull rate limit", are we past that now?  Do the /rerun commands work here?

milan-zededa:
Will we see the new workflows only after merge?

PR #1054, opened 2025-03-11
fix: create errors from non-constant strings

europaul:
Since go 1.24 vet reports the usage of non-constant strings in Printf as errors (see https://tip.golang.org/doc/go1.24#vet). This change fixes the found occurrences in our tests.

Example:
```
./nw_test.go:91:21: non-constant format string in call to fmt.Errorf
```

This fix is mostly relevant for local builds as we are not yet using go 1.24 in the CI, but the nature of the issue itself is independent of the go version used.

### DIFF
--- a/tests/app/app_test.go
+++ b/tests/app/app_test.go
@@ -1,6 +1,7 @@
 package lim
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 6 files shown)


europaul:
@uncleDecart could you please restart the failed tests? I think they are just being flaky

eriknordmark (approved):
LGTM

Issue #1028, opened 2024-09-19
QEMU version in the prerequisites page of Eden needs to be updated


roja-zededa:
# Description

Running Eden with QEMU 9.0 and QEMU 4.2.1 failed due to deprecated parameters. 

The prerequisites in the Eden says at least QEMU 4.x or higher version is required, but using QEMU 4.2.1 and 9.0.1 versions resulted in a failure during eden setup. 

Here is the log: 
`ERRO[0001] log content: qemu-system-aarch64: -chardev socket,id=char0,port=17777,host=localhost,server,nodelay,nowait,telnet,logappend=on,logfile=/Users/rojaeswaran/github.com/zededa/pr-branches/eden/dist/default-eve.log: warning: short-form boolean option 'server' deprecated
Please use server=on instead
qemu-system-aarch64: -chardev socket,id=char0,port=17777,host=localhost,server,nodelay,nowait,telnet,logappend=on,logfile=/Users/rojaeswaran/github.com/zededa/pr-branches/eden/dist/default-eve.log: warning: short-form boolean option 'nodelay' deprecated
Please use nodelay=on instead
qemu-system-aarch64: -chardev socket,id=char0,port=17777,host=localhost,server,nodelay,nowait,telnet,logappend=on,logfile=/Users/rojaeswaran/github.com/zededa/pr-branches/eden/dist/default-eve.log: warning: short-form boolean option 'nowait' deprecated
Please use wait=off instead
qemu-system-aarch64: -chardev socket,id=char0,port=17777,host=localhost,server,nodelay,nowait,telnet,logappend=on,logfile=/Users/rojaeswaran/github.com/zededa/pr-branches/eden/dist/default-eve.log: warning: short-form boolean option 'telnet' deprecated
Please use telnet=on instead
qemu_madvise: Invalid argument
madvise doesn't support MADV_DONTDUMP, but dump_guest_core=off specified
qemu_madvise: Invalid argument
madvise doesn't support MADV_DONTDUMP, but dump_guest_core=off specified
qemu_madvise: Invalid argument
madvise doesn't support MADV_DONTDUMP, but dump_guest_core=off specified
qemu-system-aarch64: -device virtio-net-pci,netdev=eth0,mac=02:fe:24:97:b1:67: failed to find romfile "efi-virtio.rom"
[...]

PR #1058, opened 2025-03-31
updating QEMU parameters and default path for the firmware

AlissonMachadoEmbitel:
Hello,

I was facing a couple of issues when trying to run it on MacOS, then I made these fixes.

It was the same problem related on this issue: https://github.com/lf-edge/eden/issues/1028

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -123,7 +123,7 @@ const (
 	DefaultAdamLogLevel = "info" // min level of logs sent from EVE to Adam
 
 	DefaultQemuAccelDarwin      = "-machine q35,accel=hvf -cpu kvm64,kvmclock=off "
-	DefaultQemuAccelDarwinArm64 = "-machine virt,accel=hvf,usb=off,dump-guest-core=off -cpu host "
+	DefaultQemuAccelDarwinArm64 = "-machine virt,accel=hvf,usb=off,dump-guest-core=on -cpu host "
 	DefaultQemuAccelLinuxAmd64  = "-machine q35,accel=kvm,dump-guest-core=off,kernel-irqchip=split -cpu host,invtsc=on,kvmclock=off -device intel-iommu,intremap=on,caching-mode=on,aw-bits=48 "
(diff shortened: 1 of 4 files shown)


uncleDecart:
Hey @AlissonMachadoEmbitel thank you very much for your contribution! Couple of things:

1. We use commit signing in, so all of our commits are signed by authors, you can easily do that by running 
```bash
git commit --amend  --signoff
```

2. Could you please tell me what qemu version are you using?
You can do it by running if you installed it with brew or you have qemu binary in your PATH 
```bash 
qemu-system-x86_64 --version
```

AlissonMachadoEmbitel:
Hi @uncleDecart ,

This is my current version:

➜  ~ qemu-system-x86_64 --version
QEMU emulator version 9.2.2
Copyright (c) 2003-2024 Fabrice Bellard and the QEMU Project developers

europaul:
I think it's a valid concern and I've seen this issue before, but it works on my combo QEMU 10.0.2 on Linux.
@shjala did you face something similar on macOS?

pkg/defaults/defaults.go
```diff
@@ -123,7 +123,7 @@ const (
 	DefaultAdamLogLevel = "info" // min level of logs sent from EVE to Adam
 
 	DefaultQemuAccelDarwin      = "-machine q35,accel=hvf -cpu kvm64,kvmclock=off "
-	DefaultQemuAccelDarwinArm64 = "-machine virt,accel=hvf,usb=off,dump-guest-core=off -cpu host "
+	DefaultQemuAccelDarwinArm64 = "-machine virt,accel=hvf,usb=off,dump-guest-core=on -cpu host "
```
rene, line 126: This option is not needed for macOS. This just includes the guest memory in a core dump. The default is already `on`, but let's keep it off to be consistent with the options for Linux.

rene:
I tested on macOS, it works. @AlissonMachadoEmbitel , I know it's been a year from this PR, but if you are still interested, I just have one comment to be addressed and we are good to merge....

PR #1065, opened 2025-07-03
Update Dockerhub token

rucoder:
- We need to use a pro version of dockerhub token to avoid "429 Too Many Requests" error
- Also we need to login to dockerhub before building or we can reach "unauthenticated limit reached" error


deitch (approved):
Looks good to me, but I am not the right one to review this. Someone like Yash/Ruslan/Nikolay should review this.

PR #1069, opened 2025-07-07
Add conditional DockerHub login to all test jobs.

OhmSpectator:
Introduce a DockerHub login step across all relevant jobs in the test workflow. This login is conditionally executed only when the workflow runs in the `lf-edge/eve` repository. It uses the official docker/login-action@v3 and authenticates with secrets already available in the repository EVE repository.

This change ensures that Docker image pulls are authenticated in trusted environments, which can help avoid rate limits or access restrictions when using private or rate-limited DockerHub resources.


uncleDecart (approved):
LGTM, we still need to do login in docker lib inside eden repo

PR #1071, opened 2025-07-09
Print public IP for buildjet runners

uncleDecart:
Apparently, DockerHub blacklisted some of IPs, we need to tell them explicitly which ones, so we print them during workflow

### DIFF
--- a/.github/actions/run-eden-test/action.yml
+++ b/.github/actions/run-eden-test/action.yml
@@ -42,6 +42,12 @@ inputs:
 runs:
   using: 'composite'
   steps:
+    - name: Print Public IP Address
+      run: |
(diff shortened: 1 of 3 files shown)


.github/actions/run-eden-test/action.yml
```diff
@@ -42,6 +42,11 @@ inputs:
 runs:
   using: 'composite'
   steps:
+    - name: Print Public IP Address
+      run: |
+        echo "Public IP Address of the runner:"
+        curl -s https://api.ipify.org
```
OhmSpectator, line 48: Are we sure curl is available on all the runners? Shouldn't we check/install?

OhmSpectator:
Rebase please and let's check if it works

PR #1072, opened 2025-07-09
Get docker registry credentials from environment variables

shjala:
This is a quick 'n dirty fix for the docker rate limiting issue, get the docker registry credentials from environment variables.

### DIFF
--- a/pkg/expect/docker.go
+++ b/pkg/expect/docker.go
@@ -3,6 +3,7 @@ package expect
 import (
 	"bytes"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/google/go-containerregistry/pkg/crane"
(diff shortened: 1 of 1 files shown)


pkg/expect/docker.go
```diff
@@ -73,11 +74,13 @@ func (exp *AppExpectation) checkDataStoreDocker(ds *config.DatastoreConfig) bool
...
-		DType:      config.DsType_DsContainerRegistry,
-		Fqdn:       exp.getDataStoreFQDN(true),
-		ApiKey:     "",
-		Password:   "",
+		Id:    id.String(),
+		DType: config.DsType_DsContainerRegistry,
+		Fqdn:  exp.getDataStoreFQDN(true),
+		// Quick fix for docker registry auth!
+		// TODO : get this value either from eden config or pod config command
+		ApiKey:     strings.TrimSpace(os.Getenv("DS_DOCKER_API_KEY")),
```
uncleDecart, line 82: that will panic if there's no ENV variable, no?

shjala: Getenv should return empty string

PR #1073, opened 2025-07-09
docker.go: Fetch and setup registry credentials

rene:
## Description
Fetch and setup registry datastore local credentials (when available), so pull operations can be performed authenticated.

## Notes
I've tested it only locally and it works, credentials from ~/.docker/config.json are retrieved correctly.

### DIFF
--- a/pkg/expect/docker.go
+++ b/pkg/expect/docker.go
@@ -90,6 +90,15 @@ func (exp *AppExpectation) createDataStoreDocker(id uuid.UUID) *config.Datastore
 		}
 	}
 
+	fmt.Println("DEBUG: fqdn:", fqdn)
+	fmt.Println("DEBUG: fqdn-true:", exp.getDataStoreFQDN(true))
+	fmt.Println("DEBUG: docker credentials:", username)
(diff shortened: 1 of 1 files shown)


pkg/expect/docker.go
```diff
@@ -72,12 +73,30 @@ func (exp *AppExpectation) checkDataStoreDocker(ds *config.DatastoreConfig) bool
...
+
+	// Return the new datastore
 	return &config.DatastoreConfig{
 		Id:         id.String(),
 		DType:      config.DsType_DsContainerRegistry,
-		Fqdn:       exp.getDataStoreFQDN(true),
-		ApiKey:     "",
-		Password:   "",
+		Fqdn:       fqdn,
+		ApiKey:     username,
```
shjala, line 98: ApiKey is username?

shjala: seems it is https://github.com/lf-edge/eve/blob/dfb85f6653a4c6069fba3a94f316741b7d856d99/pkg/pillar/cmd/downloader/resolveconfig.go#L227

PR #1077, opened 2025-07-15
Embed prebuilt eden-sdn VM image inside the container

milan-zededa:
Previously, the `lfedge/eden-sdn` container included only the SDN "service" layer (`sdnagent` and its dependencies). This change builds a complete eden-sdn VM imag -- including bootloader, kernel, containerd, etc. -- and packages it directly into the container, similar to how EVE images are handled in `lfedge/eve`.

This significantly speeds up the `eden-sdn` setup process by eliminating the need to rebuild the VM image during each `eden setup`.

Additionally, the VM image now uses legacy BIOS boot instead of UEFI, which is perfectly sufficient for `eden-sdn` and simplifies the boot process.

Finally, `eden-sdn` now follows its own semantic versioning, independent of eden. Its version is manually defined in `sdn/VERSION`, ensuring that `eden-sdn` is not unnecessarily rebuilt and published with a new image tag when unrelated eden components change.

These are initial steps towards integrating `eden-sdn` into eden tests and github workflows.


pkg/defaults/defaults.go
```diff
@@ -59,6 +59,7 @@ const (
 	DefaultRegistryTag          = "2.7"
 	DefaultProcTag              = "83cfe07"
 	DefaultMkimageTag           = "8.5.0"
+	DefaultSDNVersion           = "v1.0.0"
```
uncleDecart, line 62: can you also add this [here](https://github.com/lf-edge/eden/blob/00d2e6c817108530efe3dea4a1eaa2b5d47a0382/pkg/openevec/defaults.go#L17)?

uncleDecart:
LGTM, minor comments, could you also please add in SDN readme how can I build and push image to my local registry?

pkg/defaults/defaults.go
milan-zededa: It is already there: https://github.com/milan-zededa/eden/blob/sdn-vm-image/pkg/openevec/defaults.go#L188

uncleDecart: Oh, missed that, thank you!

milan-zededa:
@uncleDecart Checks completed and are all green...

uncleDecart:
> @uncleDecart Checks completed and they are all green...

I see you're pressuring me into merging this :D, fair enough, LGTM

uncleDecart (approved):

PR #1078, opened 2025-07-16
Add disable logs option for pod configuration

europaul:
This commit introduces a new option to disable upload of logs from pods to Adam.


europaul:
@uncleDecart yes, I used it to check a bug, but I don't have the time to write a test right now

PR #1080, opened 2025-07-17
Print public IP for buildjet runners

uncleDecart:
Apparently, DockerHub blacklisted some of IPs, we need to tell them explicitly which ones, so we print them during workflow


milan-zededa (approved):

OhmSpectator:
I'm trying to find the added line ("Public IP Address") in the tests started here, and I cannot... Strange...

uncleDecart:
> I'm trying to find the added line ("Public IP Address") in the tests started here, and I cannot... Strange...

My guess is new workflow version is not picked, not sure why, because we checkout the commit head in action

PR #1083, opened 2025-07-18
Vector integration tests

europaul:
Here I added some end-to-end tests for the vector package. The tests include:
- applying a faulty config and seeing how vector recovers
- applying an empty config and checking if vector restores the default one
- applying a working config and checking it's effects

This PR can be merged when an EVE version that includes a vector package is release and set as default for Eden.


OhmSpectator:
Do you think we should also add the new tests to the Eden workflow?

europaul:
yes, I think it should be done when EVE with vector is released.
but actually they rely a lot on SSHing into EVE and since this is the most flaky part of our Eden testing I expect they will fail a lot

OhmSpectator:
Part of  which Eden job will it be?

OhmSpectator:
Ok, why is it draft? Can we merge it?

europaul:
> Ok, why is it draft? Can we merge it?

This PR can be merged when an EVE version that includes a vector package is released and set as default for Eden.

OhmSpectator:
I think we can still merge it and just update the Eden version in EVE later.

uncleDecart:
> I think we can still merge it and just update the Eden version in EVE later.

Then we will have broken test in Eden workflows

europaul:
> @europaul we can run those tests as `go test`, right?

yap. And in a dozen other different ways that we support :laughing:

europaul:
> @europaul could you please also add 15.8.0 [here](https://github.com/lf-edge/eden/blob/4ff06a1bc50860cdf73864d74420f754cd3414d3/.github/workflows/eden.yml#L18) as well?

Done

PR #1085, opened 2025-07-21
Remove continue-on-error so job failures fail the workflow.

OhmSpectator:
The smoke, storage and eve-upgrade jobs no longer ignore step errors, ensuring any failing job marks the whole workflow as failed. To keep all matrix permutations running, each job’s strategy now explicitly sets fail-fast: false. This preserves full test coverage without cancelling siblings early while still surfacing problems in red at the workflow level.


uncleDecart:
The initial idea of that was that it takes us multiple runs to make them work, do you think that eden tests are stable enough to enable that?

europaul (approved):

uncleDecart (approved):

PR #1087, opened 2025-07-22
Skip already passed tests using artifact markers.

OhmSpectator:
Introduce logic to skip Eden jobs that have already passed by using artifacts as persistent markers. This allows using the "Rerun all jobs" button in GitHub Actions without rerunning successful tests.

We currently report commit status manually in a parent workflow after this one completes. If we use "Rerun failed jobs", GitHub delays triggering that status step until all re-executed jobs are finished. This causes a stale UI with outdated results until the rerun completes.

To refresh the UI promptly, we need to use "Rerun all jobs". With this change, already passed tests are skipped early, preventing wasted time while still ensuring the status is updated.

Each job checks for a previously uploaded `passed-marker` artifact. If present, the job exits early. On success, it re-uploads the marker to allow future skips.


.github/workflows/test.yml
```diff
@@ -68,19 +68,41 @@ jobs:
...
+            echo "already_passed=false" >> "$GITHUB_OUTPUT"
+          fi
+      - name: Early exit if already passed
+        if: steps.test_passed.outputs.already_passed == 'true'
+        run: |
+          echo "Test for $GITHUB_SHA ${{ github.job }} (${{ matrix.file_system }}, tpm=${{ matrix.tpm }}) already passed. Skipping."
+          exit 0
       - name: Dockerhub Login
-        if: ${{ github.event.repository.full_name }} == 'lf-edge/eve'
+        if: ${{ github.event.repository.full_name == 'lf-edge/eve' && steps.test_passed.outputs.already_passed != 'true' }}
```
uncleDecart, line 92: why comparison if we have early exit?

OhmSpectator: @uncleDecart, ah, the problem that exit 0 only stops the step, but not the entire job...

OhmSpectator: Removed the unnecessary early exits.

OhmSpectator:
I'll test it in Rene's repo first: https://github.com/rene/eden/pull/1

OhmSpectator:
No, it does not work as expected...

OhmSpectator:
It works, tested in Renes repo!
@uncleDecart take a look please

PR #1094, opened 2025-08-11
GHA: retire GCP

uncleDecart:
switching GCP to manual use, since they've not been up to date


OhmSpectator (commented):
I would just remove GCP completely...

PR #1097, opened 2025-09-24
eclient: add LPS network endpoint

milan-zededa:
Implement the recently introduced LPS `/api/v1/network` endpoint in `local_manager` (part of the `eclient` application) to support local network configuration changes.

The endpoint:
- Accepts `NetworkInfo` from EVE, saves it into `/mnt/network-info.json`.
- Optionally reads and returns `LocalNetworkConfig` from `/mnt/local-network-config.json` if present.
- Handles empty config file as a throttling request.

The formal definition of the LPS network endpoint can be found here: https://github.com/lf-edge/eve-api/blob/main/PROFILE.md#network

This will be used in a new LPS test for the network endpoint. 
However, this PR must be merged first so that a new version of `eclient` is built and pushed to Docker Hub, 
which the test will then reference.

### DIFF
--- a/tests/eclient/image/Dockerfile
+++ b/tests/eclient/image/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.18-alpine AS build
+FROM golang:1.23-alpine AS build
 
 WORKDIR /go/src/local_manager
 COPY pkg .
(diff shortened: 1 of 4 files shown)


uncleDecart (approved):
LGTM, weird Yetus comments

PR #1098, opened 2025-09-25
Add test for LPS /api/v1/network endpoint

milan-zededa:
This test validates local network configuration changes via the [LPS network endpoint](https://github.com/lf-edge/eve-api/blob/main/PROFILE.md#network) in the `eclient/local_manager` application.
It covers:

- Enabling local modifications on specific adapters (`eth0`/`eth1`)
- Applying local MTU and DNS overrides (as an example of a local config change)
- Verifying that modifications are rejected when disallowed by controller
- Confirming local changes are correctly applied when allowed
- Reverting local changes to controller-provided configuration

I used escript and not neoeden since the existing LPS test suite is written in escript and I'm not sure if the two can be combined under the same workflow.

For now this PR is Draft -- the EVE-side implementation of the LPS network endpoint is not yet merged and released (PR with part I is open here: https://github.com/lf-edge/eve/pull/5237; part II will follow)

### DIFF
--- a/.github/workflows/eden.yml
+++ b/.github/workflows/eden.yml
@@ -19,7 +19,7 @@ jobs:
     name: Execute Eden test workflow
     uses: ./.github/workflows/test.yml
     with:
-      eve_image: "lfedge/eve:15.8.0"
+      eve_image: "lfedge/eve:15.11.0"
       eve_kubevirt_image: "lfedge/eve:0.0.0-master-75241279-kubevirt-amd64"
       eden_version: ${{ github.event.pull_request.head.sha }}
     secrets: inherit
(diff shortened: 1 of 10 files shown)


milan-zededa:
@uncleDecart I can’t confirm whether the new test is actually passing. GitHub Actions are still running with EVE version 15.8.0, even though I updated it to 15.11.0. I’m not sure if something has changed in the workflows regarding how the EVE version is selected, but it seems that the update will only take effect after the PR is merged.

milan-zededa:
@uncleDecart We can also merge this one. Only then we will see if the test actually passes (locally it does).

uncleDecart (approved):

PR #1100, opened 2025-10-07
neoeden: propagate flags to Go tests

uncleDecart:
Since neoeden is an all-in-one solution written in Go, we now handle test setup directly within neoeden instead of relying on the setup-environment action. Previously, that action generated configuration via `eden config`, set environment parameters (e.g. TPM vs. no TPM, ZFS vs. ext4), and prepared the test environment.

With this change, neoeden performs setup in-house by passing flags directly to `go test`. This commit introduces `framework.go` (working title), which encapsulates setup logic and provides decorators to configure tests based on their required flags and parameters.


uncleDecart:
debugging neoeden tests rn..

PR #1101, opened 2025-10-07
Stop reading network info from deprecated ZInfoDevice.Network

milan-zededa:
[ZInfoDevice.Network](https://github.com/lf-edge/eve-api/blob/main/proto/info/info.proto#L596-L599) (ZInfoNetwork) has been obsolete for several years and [will soon no longer be published by EVE](https://github.com/lf-edge/eve/pull/5283). Network-related information is instead provided through [ZInfoDevice.SystemAdapter](https://github.com/lf-edge/eve-api/blob/main/proto/info/info.proto#L615) (SystemAdapterInfo), which includes additional and more up-to-date fields such as configuration source, MTU, domain name, etc.


milan-zededa:
> @milan-zededa are we okay with merge?

I think so, tests looks OK.

PR #1118, opened 2026-03-05
eclient: fix ctrl_cert_change test for EVE persist2memory branch

eriknordmark:
## Summary

Fixes one issue in `tests/eclient/testdata/ctrl_cert_change.txt` needed to handle the EVE changes where signing certs are no longer persisted to `/persist/certs/server-signing-cert.pem` and are instead kept in memory (checkpointed to `/persist/checkpoint/controllercerts`).

**`check_sign_cert.sh`**:  In older versions the signing cert was stored in a PEM file in /persist/certs/, and also in a JSON file in /persist/status/zedagent/ControllerCert/. In newer EVE versions it is available as a JSON file in /run/zedagent/ControllerCert/. We look in the new directory (/run/zedagent) and fallback to the old (/persist/status/zedagent). The signing cert has Type=1 (CERT_TYPE_CONTROLLER_SIGNING) and the PEM-encoded cert bytes are base64-encoded in the "Cert" field.

## Test plan

- [X] Run `TestEdenScripts/ctrl_cert_change` against  EVE master
- [ ] Run `TestEdenScripts/ctrl_cert_change` against an EVE build from the persist2memory branch (post merge)
- [X] Confirm `check_sign_cert.sh` correctly extracts and compares the signing cert from `/run/zedagent/ControllerCerts/`

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -82,7 +82,7 @@ const (
 	DefaultEServerTag          = "4b71e2c"
 	DefaultEServerContainerRef = "lfedge/eden-http-server"
 
-	DefaultEClientTag          = "070b10b"
+	DefaultEClientTag          = "7a72275"
 	DefaultEClientContainerRef = "lfedge/eden-eclient"
 
 	//DefaultRepeatCount is repeat count for requests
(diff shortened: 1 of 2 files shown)


uncleDecart (approved):
I'm guessing this is first contribution with Claude Code here, LGTM

PR #1134, opened 2026-04-14
Change default EVE disk image size to 16GB

rene:
This commit updates the EVE default image size from 8GB to 16GB. This increase is needed for future usage, eve-k already uses a bigger partition layout and eve-kvm will eventually be updated as well.


europaul (approved):
Yes, definitely needs to be done!

eriknordmark (approved):
LGTM

europaul:
@rene please don't forget to backport this change

PR #1135, opened 2026-04-15
Increase disk size on ZFS Tests from GH action

rene:
Commit cd853ca4d4c360c4a7d2658d622c275091e8eaaa changed the default disk size from 4GB to 16GB. However, ZFS tests have a specific setup on GH actions. This commit changes the disk size on this setup from 4GB to 16GB as well.


eriknordmark (approved):
LGTM

europaul:
@rene please don't forget to backport this change

PR #1142, opened 2026-04-17
fix: skip pre-reboot buffered logs in TestLogLevelsDifferent

rene:
# Description

This is a trial to fix the flakiness of smoke tests.

Root cause
----------
TestLogLevelsDifferent (eden_newlogd smoke test) was flaky with a false failure: "Logs found, but they should not be present with the remote log level 'none'".

The test flow is:
1. Set remote.loglevel=none on EVE.
2. WaitForConfigApplied polls EVE via SSH (stat -c %Y /persist/checkpoint) until the checkpoint file changes.
3. Each SSH poll session that disconnects writes a "Received disconnect" entry into EVE's sshd log — at this point the new log level is not yet applied on EVE, so newlogd persists those entries to disk.
4. Reboot EVE.
5. After reboot, newlogd replays all persisted pre-reboot logs to Adam.
6. `FindLogOnAdam(LogNew)` calls `XRead($)` on the Redis stream, blocking for the next entry after the current stream head.  The replayed log from step 3 arrives in Redis just after `XRead($)` starts → the watcher sees it as a new, post-config-change log → test fails.

Observed in CI (Smoke ext4,false):
  content: Received disconnect from 192.168.0.2 port 56928
  timestamp: 2026-04-15 12:33:48  (generated during WaitForConfigApplied)
  arrived in Adam at 12:36:20      (replayed by newlogd post-reboot)
  XRead($) started at 12:36:19    → race lost by 1 second

Fix
---
* Add EdenFindLogsAfter (openevec/eden.go) and FindLogOnAdamAfter (evetestkit/utils.go): same as the existing Find* pair but the match handler returns false (skip, keep watching) for any log entry whose Timestamp is before the supplied 'after' time.  Only entries at or after that timestamp trigger a match.

* In TestLogLevelsDifferent, record configAppliedAt := time.Now() immediately after WaitForConfigApplied returns (the moment EVE has confirmed the new log level), then pass it to FindLogOnAdamAfter. Pre-reboot logs replayed by newlogd are silently ignored; only a log generated *after* the new config was applied counts as a violation.

### DIFF
--- a/pkg/evetestkit/utils.go
+++ b/pkg/evetestkit/utils.go
@@ -673,6 +673,13 @@ func (node *EveNode) FindLogOnAdam(query map[string]string, mode elog.LogChecker
 	return node.controller.EdenFindLogs(query, mode, timeout)
 }
 
+// FindLogOnAdamAfter queries Adam for a log entry generated at or after the given time.
+// Log entries with timestamps before 'after' are silently skipped and do not count as a match.
+// Returns nil if a matching post-'after' entry is found, or an error (e.g. timeout) if not found.
+func (node *EveNode) FindLogOnAdamAfter(query map[string]string, mode elog.LogCheckerMode, timeout time.Duration, after time.Time) error {
+	return node.controller.EdenFindLogsAfter(query, mode, timeout, after)
+}
+
 // GetDefaultVMConfig returns a default configuration for a VM
 func GetDefaultVMConfig(appName, cloudConfig string, portPub []string) openevec.PodConfig {
 	var pc openevec.PodConfig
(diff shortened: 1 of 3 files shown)


uncleDecart:
@rene did it pass locally?

rene:
> @rene did it pass locally?

@uncleDecart TBH I'm relying on the tests here in the PR, I just got one Smoke failing due to timeout, I'm re-running... at a first glance it looks better than before for me....

europaul (approved):

eriknordmark (approved):
LGTM

PR #1143, opened 2026-04-17
eclient,log_test,userdata: ssh robustness improvements

eriknordmark:
metadata test: SSH retry robustness

Replace the iteration-bounded seq/sleep loop in ssh.sh with a
time-bounded while loop (18 min, explicit exit 1 on expiry) to reliably
fill the exec -t 20m budget.

---

eclient,log_test,userdata: ssh robustness improvements

Increase SSH timeouts, add failure indication and sshd readiness check

Replace iteration-bounded seq/sleep SSH retry loops in eclient.txt and
userdata.txt with time-bounded while loops, which exit 1 if it never
manages to connect.

In log_test.txt, add a foreground wait_ssh.sh step (5m timeout) that
polls `eden eve ssh` until EVE's sshd accepts a connection before the
background ssh.sh starts generating log entries.


rene (approved):
LGTM

PR #1144, opened 2026-04-20
tests: improve robustness against log pipeline delays and CI load

eriknordmark:
## Problem

Two intermittent test failures observed in CI (ext4+TPM matrix):

1. **`TestLog` (smoke suite)** — After EVE reboots, the log delivery pipeline
   can be completely silent for 15+ minutes while startup logs drain.  Waiting
   longer or generating more SSH connections doesn't help because the disconnect
   entries are queued behind the backlog and never reach the controller within
   the test window.

2. **`switch_net_vlans`** — Under CI load, VLAN networking for app1 takes
   longer to become reachable than the 1-minute retry window in
   `wait_and_get_recv_data.sh` allowed.

### `tests/lim/testdata/log_test.txt`
Switch to a two-phase approach:
- **Phase 1** (25m timewait): wait for ANY log entry arriving at the controller,
  proving the pipeline is active and delivering fresh entries.
- **Phase 2** (7m timewait): only *then* start generating SSH connections and
  look for the "Disconnected" message — guaranteeing it won't be lost in a
  backlog.

### `tests/network/testdata/switch_net_vlans.txt`
Replace the fixed-count loop (12 × 5 s = 60 s) in `wait_and_get_recv_data.sh`
with a time-bounded loop (~4.5 min), and raise the enclosing `exec` timeout
from 5 m to 7 m.

### DIFF
--- a/tests/eclient/testdata/nginx.txt
+++ b/tests/eclient/testdata/nginx.txt
@@ -19,7 +19,7 @@ exec chmod 600 {{EdenConfig "eden.tests"}}/eclient/image/cert/id_rsa
 
 eden pod deploy -n eclient --memory=512MB {{template "eclient_image"}} -p {{template "port"}}:22
 
-eden pod deploy -n {{$server}} --memory=512MB docker://nginx:latest
+eden pod deploy -n {{$server}} --memory=512MB docker://nginx:latest -p 8080:80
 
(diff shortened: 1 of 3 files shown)


uncleDecart (approved):
LGTM

rene (approved):
LGTM

PR #1145, opened 2026-04-22
coverage: replace fixed sleep with count-based poll after SIGUSR2

eriknordmark:
## Summary

- Replace the fixed 3-second `time.Sleep` after sending SIGUSR2 to zedbox with a count-based polling loop (30 s timeout, 2 s interval).
- Before signalling, record the number of existing `covcounters.*` files on EVE (`/tmp/eve_cov_count`); then poll until that count increases — meaning a new file was written in response to *our* signal, not left over from a previous run.
- Falls back to the original fixed sleep if the count-recording command fails (e.g. intermittent SSH issue).

## Why

The fixed sleep is fragile: on a slow or heavily loaded device 3 s may not be enough, and there is no acknowledgment from `runtime/coverage.WriteCountersDir` that the write is finished.  The polling approach detects completion reliably without requiring any changes to EVE itself (no new endpoint or sentinel file from zedbox needed).

## Test plan

- [x] Build EVE with `COVER=y`, run `eden test`, call `eden collect-coverage` and confirm the log shows the poll succeeding rather than timing out.
- [x] Verify that without a coverage-instrumented EVE the 30 s timeout fires and a warning is logged, but collection continues.
- [x] Confirm that if count recording is blocked (e.g. broken SSH), the fallback sleep path is exercised.

### DIFF
--- a/pkg/openevec/coverage.go
+++ b/pkg/openevec/coverage.go
@@ -32,16 +32,21 @@ func (openEVEC *OpenEVEC) waitForEveSSH(timeout time.Duration) error {
 	}
 }
 
-// waitForCoverageFile polls GOCOVERDIR on EVE until at least one covcounters
-// file newer than refFile appears, or timeout expires.  Newer-than-refFile
-// means the file was written after our SIGUSR2, not left over from a previous
-// collection.
-func (openEVEC *OpenEVEC) waitForCoverageFile(coverDir, refFile string, timeout time.Duration) error {
+// waitForCoverageFile polls GOCOVERDIR on EVE until the covcounters file
+// count exceeds the value recorded in countRef, or timeout expires.
+// A higher count means the SIGUSR2-triggered write completed.
+//
+// We compare counts rather than timestamps because busybox find -newer uses
+// 1-second granularity; the new covcounters file typically lands within
+// milliseconds of the reference point and would be missed by a mtime check.
+func (openEVEC *OpenEVEC) waitForCoverageFile(coverDir, countRef string, timeout time.Duration) error {
 	const pollInterval = 2 * time.Second
 	deadline := time.Now().Add(timeout)
 	pollCmd := fmt.Sprintf(
-		"find %s -newer %s -name 'covcounters.*' 2>/dev/null | grep -q .",
-		coverDir, refFile)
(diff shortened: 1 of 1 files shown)


pkg/openevec/coverage.go
```diff
@@ -32,6 +32,28 @@ func (openEVEC *OpenEVEC) waitForEveSSH(timeout time.Duration) error {
 	}
 }
 
+// waitForCoverageFiles polls GOCOVERDIR on EVE until at least one .covcounters
+// file newer than refFile appears, or timeout expires.  Newer-than-refFile
+// means the file was written after our SIGUSR2, not left over from a previous
+// collection.
+func (openEVEC *OpenEVEC) waitForCoverageFiles(coverDir, refFile string, timeout time.Duration) error {
```
milan-zededa: Should this be called `waitForCoverageFile` (drop the "s"), given that this return as soon as one new coverage file appears?

milan-zededa: But if the intention was to reference covmeta + covcounters files, than the function naming makes sense.

pkg/openevec/coverage.go
```diff
@@ -32,6 +32,28 @@ func (openEVEC *OpenEVEC) waitForEveSSH(timeout time.Duration) error {
...
+	const pollInterval = 2 * time.Second
+	deadline := time.Now().Add(timeout)
+	pollCmd := fmt.Sprintf(
+		"find %s -newer %s -name '*.covcounters' 2>/dev/null | grep -q .",
```
milan-zededa: Aren't coverage files named `covcounters.<hash>.<pid>.<ts>`? Then the search pattern should be `covcounters.*`

pkg/openevec/coverage.go
eriknordmark: Done — renamed to `waitForCoverageFile` (singular). We only check for `covcounters.*` files so the plural name was misleading.

pkg/openevec/coverage.go
eriknordmark: Good catch — fixed to `covcounters.*`. Go names the files `covcounters.<hash>.<pid>.<ts>` so the original glob never matched anything.

milan-zededa (approved):

PR #1150, opened 2026-04-30
tests/eclient: fix ctrl_cert_change log pattern after ReconnectApp rename

rene:
The ctrl_cert_change smoke test was timing out waiting for the log message "Rebuilding intended global config, reasons: reconnecting app", which no longer exists.

On EVE, when ReconnectApp() in nireconciler/linux_reconciler.go was renamed to UpdateAppConn() as part of the Kubernetes CNI plugin work, the reason string passed to scheduleGlobalCfgRebuild() changed from "reconnecting app (%v)" to "updating app connectivity (%v)".

The mechanism itself still works correctly: a controller cert change invalidates cipher contexts, which triggers AppNetworkConfig re-publication, which causes doUpdateActivatedAppNetwork to call UpdateAppConn in the NI reconciler, which logs the updated reason string.

Update the LIM test pattern to match the current reason string.

An example of the timeout error: https://github.com/lf-edge/eve/actions/runs/25115269940/job/73757607638#step:4:3029

The changes on EVE side: https://github.com/lf-edge/eve/commit/415d685992a428f3e5cac17363f48e477476eb2b

### DIFF
--- a/tests/eclient/testdata/ctrl_cert_change.txt
+++ b/tests/eclient/testdata/ctrl_cert_change.txt
@@ -26,7 +26,7 @@ eden utils gen-signing-cert -o /tmp/signing-new.pem
 eden adam change-signing-cert --cert-file /tmp/signing-new.pem
 
 # wait for changes to be applied
-test eden.lim.test -test.v -timewait 15m -test.run TestLog -out content 'content:Rebuilding.intended.global.config,.reasons:.reconnecting.app'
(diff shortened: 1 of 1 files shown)


milan-zededa (approved):
This EVE change is quite old though. Was it failing all this time?

rene:
> This EVE change is quite old though. Was it failing all this time?

Well, that's been failing for quite a while now...

PR #1154, opened 2026-05-06
eden: rotate ECDSA keypair on controller signing-cert change

eriknordmark:
## Summary

Stacks on top of #1153. Turns \`gen-signing-cert\` from a cert-envelope refresh
into a real signing-key rotation, so the \`ctrl_cert_change\` smoke test
exercises the full controller-key change rather than just a cert-hash
mismatch on EVE.

Today \`GenServerCertFromPrevCertAndKey\`
([\`pkg/utils/x509.go\`](https://github.com/lf-edge/eden/blob/master/pkg/utils/x509.go))
reuses the existing \`signing-key.pem\` and only varies \`NotBefore\`/\`NotAfter\`,
so the new cert ends up with the same \`SubjectPublicKeyInfo\` as the old one.
The ECDH shared secret derived in \`calculateSymmetricKeyForEcdhAES\` is
therefore identical before and after - rotating the cert on adam changes the
cert-hash that EVE sees, but does not actually rotate the controller signing
key. An attacker holding the previous key would still be able to read freshly
re-encrypted ciphertexts.

## Changes

- **\`pkg/utils/x509.go\`**: replace \`GenServerCertFromPrevCertAndKey\` with
  \`GenServerCertWithNewKey(certPath, keyPath string)\`. Generates a fresh
  ECDSA P-256 keypair, picks a random 128-bit serial, copies the subject
  and other template fields from the existing \`signing.pem\`, signs with
  the eden root CA, and writes the cert plus the matching key (in the
  same \`EC PARAMETERS\` + \`EC PRIVATE KEY\` PEM form used by
  \`signing-key.pem\`, mode 0600) via the existing \`WriteToFiles\` helper.
- **\`cmd/certs.go\`**: \`gen-signing-cert\` gains a \`--key-out\` flag. If
  unset, the path is derived from \`--out\` by replacing the trailing
  \`.pem\` with \`-key.pem\`.
- **\`cmd/adam.go\`**: \`change-signing-cert\` gains an optional \`--key-file\`
  flag. When omitted, the historical cert-only rotation is preserved for
  any external caller.
- **\`pkg/openevec/adam.go\`**: \`ChangeSigningCert\` and \`reencryptConfigs\`
  now thread an optional new private key through. \`oldCryptoConfig\`
  keeps using the on-disk key; \`newCryptoConfig\` uses the new key when
  supplied, so the cipher contexts written for newly-encrypted blocks
  rely on a genuinely different ECDH secret. The new key is written to
  \`signing-key.pem\` _before_ the new cert is written to \`signing.pem\`,
  so adam never observes a (new cert, old key) pair.
- **\`tests/eclient/testdata/ctrl_cert_change.txt\`**: both rotations now
  pass \`--key-out\` to \`gen-signing-cert\` and \`--key-file\` to
  \`change-signing-cert\`.

## Test plan

- [x] Smoke (ext4, true) on EVE master with the upstream
      \`SenderStatusCertMiss\`-preservation fix in \`authen.go\`. Both
      rotations should complete inside the 15m TestLog windows, the
      "Rebuilding intended global config, reasons: updating app
      connectivity" log should fire after each, and \`check_sign_cert.sh\`
      should still succeed because \`/tmp/signing-new.pem\` holds the
      most recent rotation.
- [x] Smoke (zfs, true), Smoke (ext4, false), Smoke (zfs, false) for
      parity with the existing matrix.
- [x] Local: \`go build ./...\`, \`go vet ./...\`, \`gofmt -l\` on the
      touched files (all clean).

## Risks / open questions for reviewers

1. **Adam key-cache behavior.** This change is sound only if adam
   re-reads \`signing-key.pem\` from disk per request (which the
   single-key cert-rotation flow has been quietly relying on for
   \`signing.pem\`). If adam pins the key in memory at startup, the first
   TestLog after the first key rotation will time out with a signature
   verification failure on EVE rather than a cert-miss, and we will
   need to add an adam restart (or a key-reload signal) inside
   \`ChangeSigningCert\`. Filing this PR as draft so this can be
   confirmed in CI. I'm investigating the adam side separately.
2. **Exported Go signature change.** \`ChangeSigningCert\` changes from
   \`(newSignCert []byte)\` to \`(newSignCert, newSignKey []byte)\`. The
   only in-tree caller is the eden CLI itself; no other consumer should
   be affected, but a heads-up in release notes seems prudent.
3. **\`gen-signing-cert\` semantic shift.** Anyone who relied on the
   "preserve the same controller key" behavior will now silently get a
   key rotation. Flag set is back-compat (no flags removed), but the
   underlying behavior is not.

## Stacking

Built on \`ctrl-cert-change-twice\` (PR #1153). Once that lands this can
be rebased onto master; until then please review on top of #1153.

### DIFF
--- a/cmd/adam.go
+++ b/cmd/adam.go
@@ -95,25 +95,37 @@ func newStatusAdamCmd() *cobra.Command {
 }
 
 func newChangeCertCmd() *cobra.Command {
-	var certFile string
+	var certFile, keyFile string
 
 	var changeCertCmd = &cobra.Command{
 		Use:   "change-signing-cert",
 		Short: "change signing certificate for adam",
-		Long:  `Set Adam's signing certificate from a file.`,
+		Long: `Set Adam's signing certificate from a file. If --key-file is
+provided, the matching private key is also installed and used to re-encrypt
+configs; otherwise only the certificate is rotated and the existing
+signing-key.pem is reused.`,
 		Run: func(cmd *cobra.Command, args []string) {
 			certData, err := os.ReadFile(certFile)
 			if err != nil {
 				log.Fatalf("Failed to read certificate file: %s", err)
 			}
 
-			if err := openEVEC.ChangeSigningCert(certData); err != nil {
+			var keyData []byte
+			if keyFile != "" {
+				keyData, err = os.ReadFile(keyFile)
+				if err != nil {
+					log.Fatalf("Failed to read key file: %s", err)
+				}
+			}
+
+			if err := openEVEC.ChangeSigningCert(certData, keyData); err != nil {
 				log.Fatalf("Failed to upload certificate to adam: %s", err)
 			}
 		},
 	}
 
-	changeCertCmd.Flags().StringVarP(&certFile, "cert-file", "", "", "path to the signing certificate file")
+	changeCertCmd.Flags().StringVarP(&certFile, "cert-file", "", "", "path to the new signing certificate file")
+	changeCertCmd.Flags().StringVar(&keyFile, "key-file", "", "path to the new signing private key (optional; if omitted, the existing key is reused)")
 
 	return changeCertCmd
 }
--- a/cmd/certs.go
+++ b/cmd/certs.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/lf-edge/eden/pkg/defaults"
 	"github.com/lf-edge/eden/pkg/eden"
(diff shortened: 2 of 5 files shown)


eriknordmark:
**Risk #1 (adam key-cache) verified.**

Walked through `lf-edge/adam` master (commit `61f045f`). Both the cert and the private key are re-read from disk on every API response that requires them; no in-memory cache between requests:

- `pkg/server/apiHandlerv2.go:357` `prepareEnvelope()` is the function that wraps every v2 response in an `AuthContainer`. It is called from:
  - `apiHandlerv2.go:278` — `/api/v2/edgedevice/certs` (the cert-fetch path EVE hits after `SenderStatusCertMiss`)
  - `apiHandlerv2.go:510` — `/api/v2/edgedevice/.../config` (the routine config-pull path)
  - `apiHandlerv2.go:533` — `attest`
  - `apiHandlerv2.go:686` — `register/uuid`
- Inside `prepareEnvelope`:
  - `apiHandlerv2.go:364` calls `getCertChain(h.signingCertPath, ...)` which at `apiHandlerv2.go:208` does `os.ReadFile(certPath)` unconditionally — no cache.
  - `apiHandlerv2.go:376` does `os.ReadFile(h.signingKeyPath)` directly.
- `getAllCerts` (`apiHandlerv2.go:133`), used by the `/certs` endpoint, also goes through `getCertChain`, so the same fresh-from-disk read applies to the cert payload itself.

Net: when `ChangeSigningCert` writes the new `signing-key.pem` followed by `signing.pem`, adam's very next API response uses both new files. No restart needed, no reload signal needed. The "(new cert, old key)" race window is bounded only by the order of the two `os.WriteFile` calls in `pkg/openevec/adam.go`, which this PR already serializes (key first, then cert).

Marking risk #1 as resolved on the adam side. Risks #2 (Go-API signature change) and #3 (`gen-signing-cert` semantic shift) still stand for reviewer judgment.

uncleDecart (approved):
LGTM, I'm okay with the risks mentioned

PR #1178, opened 2026-05-13
tests: add onboarding e2e suite

eriknordmark:
## Summary

Adds an onboarding e2e suite under `tests/onboarding/` that exercises
the cmd/client agent in EVE against a live device booted by the
standard eden setup. Four scenarios:

- `onboarding_status` — happy path. After `eden eve onboard`, asserts
  `/persist/status/uuid` parses, the kernel hostname matches that UUID,
  and `/persist/status/hardwaremodel` is non-empty.
- `server_changes_at_runtime` — writes a deliberately-wrong controller
  URL into the CONFIG partition via `eve config mount`, reboots, and
  verifies the boot exercises cmd/client's wrong-cert-fetch path
  (`LedBlinkInvalidControllerCert` escalation). Recovers via an in-VM
  reboot since EVE cannot reach adam to receive a controller-side
  reboot directive once the bad URL is active.
- `pre_provisioned_uuid` — uses the same wrong-server lever to keep
  nim's DPC state at FailWithIPAndDNS while cmd/client's 5-second t1
  timer fires, exercising the "already have UUID + checkpoint; declare
  success" shortcut branch in `Run()`.
- `server_returns_403` — drives the un-onboard flow: removes the
  device's adam-side onboard cert + device record via the admin
  endpoints, wipes `/persist/status/uuid` + `/persist/checkpoint/controllercerts`
  on EVE, reboots, asserts `LedBlinkOnboardingFailureNotFound` (=18),
  POSTs the onboard cert back, reboots again, asserts re-onboarding
  succeeds. Requires lf-edge/adam ≥ 0.0.79 (the first release that
  includes lf-edge/adam#155, which brings `/register` in line with
  the eve-api spec: 403 for cert/serial that isn't pre-registered).
  Paired with lf-edge/eden#1180 which bumps `DefaultAdamTag` so the
  default eden setup pulls the right adam.

Complements the cmd/client Go unit tests in lf-edge/eve#5957 and
lf-edge/eve#5958. Combined with the unit tests this suite raised
`pkg/pillar/cmd/client` block coverage from 0 % to **87.36 %**
(eden e2e alone gets 71.18 %; unit-only is 53.66 %).

Two further scenarios from the original plan were left out of this
PR; both need an un-onboarding harness that's now built into
`server_returns_403`, so they become follow-ups:

- `wrong_server_url` (un-onboarded variant) — overlaps with
  `server_changes_at_runtime` so the marginal coverage is small.
- `register_then_getuuid_interleave` — needs adam to know the device
  UUID *but* reject the onboard cert (a contrived state).

The original `cert_miss_refetch` scenario was found unreachable from
e2e during analysis: cmd/client's `fetchCertChain` refreshes the
in-memory cert hash before `doGetUUID` ever runs, so `SenderStatusCertMiss`
cannot fire in the same `Run()`; the branch is already covered by the
unit test `TestDoGetUUID_CertMissSchedulesTimer`.

## Test plan

- [x] `eden setup && eden start && eden test ./tests/onboarding/ -v`
      from a clean eden workspace passes all four scenarios.
- [x] Same on a coverage-instrumented EVE; `eden eve collect-coverage`
      after the run shows the cmd/client coverage numbers above
      (verified locally against an EVE built with `make COVER=y`).

### DIFF
--- a/tests/onboarding/testdata/onboarding_status.txt
+++ b/tests/onboarding/testdata/onboarding_status.txt
@@ -1,0 +1,55 @@
+# Test: onboarding produces the expected on-device artifacts.
+#
+# Asserts that after onboarding the device has:
+#   * /persist/status/uuid containing a parseable UUID,
+#   * a kernel hostname matching that UUID,
+#   * /persist/status/hardwaremodel containing a non-empty model string.
+#
+# Exercises the post-loop publishing path in pkg/pillar/cmd/client (the
+# hostname call, the uuid and hardwaremodel file writes, and the
+# OnboardingStatus publication) plus the entire onboarding event loop.
+
+[!exec:bash] stop
+
+eden eve onboard
+stdout 'onboarded'
+
+# Wait for the artifacts; zedclient writes them shortly after the
+# controller hands back the UUID. Allow up to 3 minutes for sshd
+# readiness and the filesystem write to settle.
+exec -t 3m bash check_onboarding_artifacts.sh
+stdout 'onboarding-ok'
+
+-- check_onboarding_artifacts.sh --
+#!/bin/bash
+set -uo pipefail
+EDEN={{EdenConfig "eden.root"}}/{{EdenConfig "eden.bin-dist"}}/{{EdenConfig "eden.eden-bin"}}
+
+uuid_re='^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$'
+
+uuid=""
+hn=""
+hw=""
+
+for _ in $(seq 1 30); do
+    uuid=$($EDEN eve ssh -- cat /persist/status/uuid 2>/dev/null | tr -d '\r\n ')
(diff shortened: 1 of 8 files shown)


tests/onboarding/testdata/onboarding_status.txt
```diff
@@ -0,0 +1,55 @@
...
+# Exercises the post-loop publishing path in pkg/pillar/cmd/client (the
+# hostname call, the uuid and hardwaremodel file writes, and the
+# OnboardingStatus publication) plus the entire onboarding event loop.
+
+[!exec:bash] stop
+
+eden eve onboard
+stdout 'onboarded'
+
+# Wait for the artefacts; zedclient writes them shortly after the
```
rene: `artefacts` -> `artifacts`

rene (approved):
LGTM other than the typo to be fixed

tests/onboarding/testdata/onboarding_status.txt
eriknordmark: Apparently Anthropic has been inflitrated by some Brit (but I haven't seen other odd spelling) ;-)

PR #1186, opened 2026-05-20
Change default disk size to support new EVE partition layout

rene:
## Description
    
EVE has bumped rootfs partition size from 4GB to 10GB. Adjust default disk size to support the new layout.

For reference: https://github.com/lf-edge/eve/pull/5973


eriknordmark (approved):
LGTM

PR #1196, opened 2026-06-11
[WIP] tests/update_eve_image: cross-HV (kvm↔k) BaseOs upgrade test suite

eriknordmark:
# Description

End-to-end eden tests for the EVE-kvm ↔ EVE-k cross-flavor BaseOs
upgrade, including the in-field boot-disk repartition (small → large
GPT geometry). They verify that OCI blobs are reused across the flavor
switch (no re-download), that a deployed app and its volume survive the
conversion and a controller-initiated reboot, and — for the resize
variants — that the partitions actually grow and the TPM-sealed vault
stays locally unlocked across the geometry change.

All scenarios live under `tests/update_eve_image/testdata/`, each a
self-contained escript (plus helper scripts and READMEs):

- `update_eve_image_cross_hv.txt` — base cross-HV upgrade; IMGB reaches
  `active` on the alternate flavor.
- `update_eve_image_cross_hv_with_contenttree.txt` — pre-stages a
  ContentTree, asserts zero bytes re-downloaded after the switch.
- `update_eve_image_cross_hv_with_app_recreate.txt` — app + volume
  survive the switch and a reboot with no re-download.
- `update_eve_image_kvm_to_k{,_grow,_resize,_refused,_volmig}.txt` —
  repartition variants asserting SMALL→LARGE partition growth,
  `BaseOsStatus.Converting` device state, blob reuse, local vault unlock
  (`VaultStatus.UnlockMethod`), and (volmig) live app-volume migration
  to a Longhorn PVC.
- `update_eve_image_kvm_to_k_geom.txt` — a **geometry-only** matrix,
  parameterized by the *starting release*, asserting the boot disk ends
  at the full EVE-k target **2+2+10+10** (ESP-A 2G, the reserved ESP-B 2G
  at GPT #7, IMGA/IMGB 10G) — including *creating* the reserved ESP-B
  when the start image lacks it. One escript covers the range of
  boot-disk geometries a fielded device may carry (10.1.0 at 36M/300M,
  12.1.0 at 36M/512M, 16.13.0 at 2G/4G, 17.0.0-rc1 at 2G/10G). It checks
  geometry only (no app/content-tree/vault-seal), so it needs no TPM or
  `/config` settle; the two ESPs share the `EFI System` label and are
  told apart by PARTUUID. All four starts validated GREEN (see
  `README_kvm_to_k_geom.md`).

## Dependency

These exercise EVE-side behavior that is not yet in eden's CI EVE image.
**Do not enable in CI/CD until lf-edge/eve#6036 (kvm↔k BaseOs upgrade
with blob reuse) merges**; the resize/volmig variants additionally need
an EVE build carrying the boot-disk repartition work, and the `geom`
variant needs a build whose `storage-resizer` *creates* the reserved
ESP-B (the ESP-B retrofit). Kept as a **draft** until then.

### DIFF
--- a/tests/update_eve_image/eden.update_eve_image.tests.txt
+++ b/tests/update_eve_image/eden.update_eve_image.tests.txt
@@ -1,2 +1,5 @@
 eden.escript.test -test.run TestEdenScripts/update_eve_image_http -test.timeout 60m
 eden.escript.test -test.run TestEdenScripts/update_eve_image_oci -test.timeout 60m
+eden.escript.test -test.run TestEdenScripts/update_eve_image_cross_hv -test.timeout 90m
+eden.escript.test -test.run TestEdenScripts/update_eve_image_cross_hv_with_contenttree -test.timeout 90m
+eden.escript.test -test.run TestEdenScripts/update_eve_image_cross_hv_with_app_recreate -test.timeout 180m
--- a/tests/update_eve_image/testdata/README_kvm_to_k_geom.md
+++ b/tests/update_eve_image/testdata/README_kvm_to_k_geom.md
@@ -1,0 +1,111 @@
+# kvm->k geometry matrix (`update_eve_image_kvm_to_k_geom`)
+
+A geometry-only sibling of `update_eve_image_kvm_to_k_resize`. It proves that,
+whatever the boot disk's **starting** GPT layout, driving an EVE-kvm -> EVE-k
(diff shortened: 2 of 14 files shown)


eriknordmark:
Superseded by #1209, which folds this branch's tests into a single branch (`kvm-to-k-eden-tests`) rebased onto current master.

PR #1199, opened 2026-07-01
Adapt network_local_changes DNS check for mgmt dnsmasq

milan-zededa:
Since https://github.com/lf-edge/eve/pull/6021, EVE runs a management dnsmasq on `127.0.0.1:53` and writes only `nameserver 127.0.0.1` to `/etc/resolv.conf`; upstream DNS servers are listed in `/run/nim/dnsmasq.mgmt.servers` instead.

Replace bare `cat /etc/resolv.conf` with a combined cat of both files (`2>/dev/null` suppresses the missing-file error on older EVE) so the test passes on both old and new EVE. Define a template for the command to avoid repetition. Drop the `nameserver ` prefix from stdout matchers since the servers file uses a different format.


eriknordmark (approved):
LGTM

PR #1202, opened 2026-07-02
openevec: add-wireless for encrypted WiFi port

eriknordmark:
**Motivation:** testing the kvm-to-k partition-resizing recovery from a lost /persist.

A new `eden controller edge-node add-wireless` command injects a WiFi device
port into the device config with its credentials encrypted (ECDH) against the
device certificate, so EVE decrypts them at device-config ingest. This
exercises credential decryption independent of the app pipeline and of physical
radio presence.

The encryption is factored out of the app-userdata path as a reusable
`EncryptForDevice` helper (hard error, no plaintext fallback). Since the QEMU
device model has no radio, the command synthesizes a wireless PhysicalIO and a
non-management SystemAdapter; verified in a sandbox that nim decrypts the
credentials at ingest even with no radio present.


rene (approved):
LGTM

PR #1212, opened 2026-07-11
tests/nim: split NIM e2e tests into a dedicated suite

eriknordmark:
**Draft / WIP.**

Splits the NIM end-to-end tests into a dedicated `tests/nim/` suite:
moves the nine `nim_*.txt` testdata scripts out of
`tests/network/testdata/` into `tests/nim/testdata/`, with a matching
Makefile, `eden-config.yml`, and `eden.nim.tests.txt`, so NIM-focused
coverage is reported independently of the broader network suite. The
escript runner handles them as `TestEdenScripts/<name>`, mirroring the
nodeagent and baseosmgr suites.

The LPS and radio-silence NIM tests under `tests/eclient/testdata/`
stay in the eclient suite — they depend on the eclient app and
local-manager harness, so they are NIM tests by exercise but not by
runtime dependency.


eriknordmark:
Rebased on master to get fresh CI. The red Smoke job was the `eden_vector` escript, where every subtest failed at `ssh eve status | grep vector` with `exit status 255` — EVE was not reachable over SSH on that runner. This PR only adds `tests/nim/`, so nothing in it reaches that path.

PR #1215, opened 2026-07-16
tests: add pciback suite for port kept in host on name mismatch

eriknordmark:
## Description

Adds an evetestkit suite (`tests/pciback`) that verifies domainmgr keeps a
network port in the host instead of reserving its PCI device to pciback when
the controller's device model gives the port an interface name that differs
from the kernel-assigned name (e.g. `enpNsN` vs `ethN`).

The test resolves a port's PCI address from the running IoBundle, rewrites the
device model through the controller so that the port's model interface name no
longer matches the kernel, restarts EVE, and asserts the affected NIC stays
`KeepInHost` and out of pciback while the device remains reachable.

It guards the EVE fix in lf-edge/eve#6169 (match device ports to their PCI
device, not just by interface name). Validated on the ZedVirtual-4G QEMU model:

- **passes** against an EVE build with the fix;
- **fails** against master without it (the in-use NIC is reserved to pciback,
  `IsPCIBack=true`).

## How to run

```
make -C tests/pciback build
./eden test tests/pciback -v debug
```

## Notes

- Requires the ZedVirtual-4G (QEMU) model — it depends on the two virtio NICs.
  `eth0` is left untouched so EVE stays reachable while `eth1` is the subject of
  the mismatch.
- Not yet wired into a `tests/workflow` CI scenario (the model mutation + node
  restart cycle takes several minutes). Can add it to `networking.tests.txt` if
  wanted.
- Draft until lf-edge/eve#6169 merges, since the suite only passes with that fix.

### DIFF
--- a/tests/pciback/Makefile
+++ b/tests/pciback/Makefile
@@ -1,0 +1,77 @@
+DEBUG ?= "debug"
+
+# HOSTARCH is the host architecture
+# ARCH is the target architecture
+# we need to keep track of them separately
+HOSTARCH ?= $(shell uname -m)
+HOSTOS ?= $(shell uname -s | tr A-Z a-z)
+
+# canonicalized names for host architecture
+override HOSTARCH := $(subst aarch64,arm64,$(subst x86_64,amd64,$(HOSTARCH)))
+
+# unless otherwise set, I am building for my own architecture, i.e. not cross-compiling
+# and for my OS
+ARCH ?= $(HOSTARCH)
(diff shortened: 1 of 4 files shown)


eriknordmark:
Closing: this pciback coverage is superseded by the equivalent test now implemented in eve's evetest framework, so maintaining a parallel eden suite is redundant. The underlying EVE fix is tracked in lf-edge/eve#6169.

